PR #1, opened 2015-09-17
Initial commit

illicitonion:
FYI: I have never written Node before, it feels like I should probably pull this out into at least two files, and I have no idea how to write tests for this thing, but I'd like to.

This is Kegan's HOWTO example, made configurable

It is woefully insufficient for real use, and in particular requires you
to both hard-code your list of channels in advance, and somehow know the
slack internal channel ID of the channels you are bridging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@ To install, you will need to get dependencies by running:
 $ npm init
 $ npm install matrix-appservice-bridge
 $ npm install request
-$ npm install yamljs
 ```
 
 Then fill out a config.yaml file according to the example in
@@ -22,5 +21,5 @@ $ node app.js -r -c config.yaml -u "http://localhost:9000"
 
 Start your application service:
 ```
-$ node app.js -p 9001 -c config.yaml
+$ node app.js -p 9000 -c config.yaml
 ```
--- a/README.md
+++ b/README.md
@@ -1,2 +1,26 @@
 # matrix-appservice-slack
 A Matrix &lt;--> Slack bridge
+
+This is currently a very barebones bridge, it just does basic text in
+pre-enumerated channels. It will become more exciting.
+
+To install, you will need to get dependencies by running:
+```
+$ npm init
+$ npm install matrix-appservice-bridge
+$ npm install request
+$ npm install yamljs
+```
+
+Then fill out a config.yaml file according to the example in
+config/config.sample.yaml and run the following commands:
+
+Register your application service with your homeserver:
+```
+$ node app.js -r -c config.yaml -u "http://localhost:9000"
+```
+
+Start your application service:
+```
+$ node app.js -p 9001 -c config.yaml
+```
--- a/app.js
+++ b/app.js
@@ -31,7 +31,16 @@ function startServer(config, rooms, callback) {
             if (params.user_id !== "USLACKBOT") {
                 var intent = bridge.getIntent("@" + config.username_prefix + params.user_name + ":" + config.homeserver.server_name);
                 if (rooms.knowsSlackChannel(params["channel_id"])) {
-                    intent.sendText(rooms.matrixRoomID(params["channel_id"]), params.text);
+                    var roomID = rooms.matrixRoomID(params["channel_id"]);
+                    if (roomID) {
+                        intent.sendText(roomID, params.text);
+                    } else {
+                        console.log(
+                            "Ignoring message for slack channel " +
+                            "with unknown matrix ID: " + params["channel_id"] +
+                            " (" + params["channel_name"] + ")"
+                        );
+                    }
                 }
             }
             response.writeHead(200, {"Content-Type": "application/json"});
@@ -116,13 +125,18 @@ new Cli({
 
                     onEvent: function(request, context) {
                         var event = request.getData();
-                        if (event.type !== "m.room.message" || !event.content || !rooms.knowsMatrixRoom(event.room_id)) {
+                        if (event.type !== "m.room.message" || !event.content) {
+                            return;
+                        }
+                        var hookURL = rooms.webhookForMatrixRoomID(event.room_id);
+                        if (!hookURL) {
+                            console.log("Ignoring event for matrix room with unknown slack channel:" + event.room_id);
                             return;
                         }
                         requestLib({
                             method: "POST",
                             json: true,
-                            uri: rooms.webhookForMatrixRoomID(event.room_id),
+                            uri: hookURL,
                             body: {
                                 username: event.user_id,
                                 text: event.content.body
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,2 +1,26 @@
...
+
+This is currently a very barebones bridge, it just does basic text in
+pre-enumerated channels. It will become more exciting.
+
+To install, you will need to get dependencies by running:
+```
+$ npm init
+$ npm install matrix-appservice-bridge
+$ npm install request
+$ npm install yamljs
```
Kegsay: I don't think you ever use this.

README.md
```diff
@@ -1,2 +1,26 @@
...
+config/config.sample.yaml and run the following commands:
+
+Register your application service with your homeserver:
+```
+$ node app.js -r -c config.yaml -u "http://localhost:9000"
+```
+
+Start your application service:
+```
+$ node app.js -p 9001 -c config.yaml
```
Kegsay: This should match the port when registering eg 9000

app.js
```diff
@@ -0,0 +1,137 @@
+// Usage:
+// node app.js -r -c config.yaml -u "http://localhost:9001" # remember to add the registration!
+// node app.js -p 9001 -c config.yaml
+var qs = require('querystring');
+var requestLib = require("request");
+var yaml = require("yamljs");
+var bridge;
+
+function startServer(config, rooms, callback) {
+    if ("tls" in config) {
```
Kegsay: `if (config.tls)` is cleaner.

app.js
```diff
@@ -0,0 +1,137 @@
...
+        var fs = require("fs");
```
Kegsay: `require()` is like import statements, they should really be at the top of the file.

app.js
```diff
@@ -0,0 +1,137 @@
...
+        var tls_options = {
+            key: fs.readFileSync(config["tls"]["key_file"]),
```
Kegsay: You may find `config.tls.key_file` nicer to write, and you know it's defined because of your schema.

app.js
```diff
@@ -0,0 +1,137 @@
...
+            cert: fs.readFileSync(config["tls"]["crt_file"])
+        };
+        var createServer = function(cb) { return require("https").createServer(tls_options, cb); };
+    } else {
+        var createServer = require("http").createServer;
```
Kegsay: Defining `createServer` twice like this is a little messy. Javascript will hoist `var` declarations to the top of the function, so it's generally clearer to just do the same. Rather than having it be initially `undefined` and then setting it conditionally here, you could initially assign it to this line then clobber it with the tls version if supplied.

app.js
```diff
@@ -0,0 +1,137 @@
...
+                    intent.sendText(rooms.MatrixRoomID(params["channel_id"]), params.text);
+                }
+            }
+            response.writeHead(200, {"Content-Type": "application/json"});
+            response.write(JSON.stringify({}));
+            response.end();
+        });
+    }).listen(config["slack_hook_port"], function() {
+        var protocol = "http";
+        if ("tls" in config) {
```
Kegsay: `config.tls`

app.js
```diff
@@ -0,0 +1,137 @@
...
+    this.slack_channels = {};
+    this.matrix_rooms = {};
+    for (var i = 0; i < config["rooms"].length; ++i) {
+        var room = config["rooms"][i];
+        this.slack_channels[room["slack_channel_id"]] = room;
+        this.matrix_rooms[room["matrix_room_id"]] = room
+    }
+}
+
+Rooms.prototype.KnowsSlackChannel = function(slack_channel_id) {
```
Kegsay: Method names shouldn't start with a Capital.

app.js
```diff
@@ -0,0 +1,137 @@
...
+    return (slack_channel_id in this.slack_channels);
+};
+
+Rooms.prototype.KnowsMatrixRoom = function(matrix_room_id) {
+    return (matrix_room_id in this.matrix_rooms);
```
Kegsay: Use `return Boolean(this.matrix_rooms[matrix_room_id])` to test key presence like this.

app.js
```diff
@@ -0,0 +1,137 @@
...
+};
+
+Rooms.prototype.MatrixRoomID = function(slack_channel_id) {
+    return this.slack_channels[slack_channel_id]["matrix_room_id"];
```
Kegsay: Null check is probably needed here.

Kegsay:
We'll work through this irl tomorrow. The structure is great, I just need to tell you more about Node etc

README.md
illicitonion: Done.

This was a relic from before I found out about your config magic :)

README.md
illicitonion: Oops! Done

app.js
illicitonion: Python has rotted my brain. Done.

app.js
illicitonion: But they are conditionally imported! What do I do about conditional imports??

app.js
illicitonion: Done

app.js
illicitonion: Done

(I was relying on the hoisting to make them the same variable, but I guess it is more clear to be explicit about that :))

app.js
illicitonion: Done

app.js
illicitonion: Done

app.js
illicitonion: Done

app.js
illicitonion: Done.

app.js
```diff
@@ -0,0 +1,159 @@
...
+            }
+            response.writeHead(200, {"Content-Type": "application/json"});
+            response.write(JSON.stringify({}));
+            response.end();
+        });
+    }).listen(config.slack_hook_port, function() {
+        var protocol = "http";
+        if (config.tls) {
+            protocol = "https";
+        }
```
Kegsay: You can use ternary operators here e.g.
```javascript
var protocol = config.tls ? "https" : "http";
```

app.js
```diff
@@ -0,0 +1,159 @@
...
+        console.log("Slack-side listening on port " + config.slack_port + " over " + protocol);
+        callback();
+    });
+}
+
+var Cli = require("matrix-appservice-bridge").Cli;
+var Bridge = require("matrix-appservice-bridge").Bridge;
+var AppServiceRegistration = require("matrix-appservice").AppServiceRegistration;
```
Kegsay: The newest HOW-TO code replaces the require from `"matrix-appservice"` to `"matrix-appservice-bridge"

app.js
illicitonion: Done

app.js
illicitonion: Done, also pulled out the require

Kegsay:
LGTM

PR #2, opened 2015-09-18
Pull out hook handler

illicitonion:
With tests!


Kegsay:
LGTM - Should the `HookHandler` also be handling stuff from Matrix to Slack?

illicitonion:
That's a separate class I'll send you a pull request for shortly :)

Perhaps I should really be naming HookHandler SlackHookHandler... Will see :)

PR #3, opened 2015-09-21
Pull out MatrixHandler

illicitonion:


app.js
```diff
@@ -5,6 +5,7 @@ var qs = require("querystring");
 var requestLib = require("request");
 var Rooms = require("./lib/rooms");
 var HookHandler = require("./lib/hook-handler");
```
Kegsay: As you said, given this is handling Slack requests only, the class name should be indicative of this.

Kegsay:
LGTM!

app.js
illicitonion: Added Slack to the name

PR #4, opened 2015-09-21
Perform special character substitution

illicitonion:
Covers &<>

Still maybe to do:
 * Emoji
 * Emotes
 * Font stylings
 * Anything else which comes up

### DIFF
--- a/lib/substitutions.js
+++ b/lib/substitutions.js
@@ -12,9 +12,17 @@ function add(slack, matrix) {
 
 // slack -> matrix substitutions are performed top -> bottom
 // matrix -> slack substitutions are performed bottom -> top
+//
+// The ordering here matters because some characters are present in both the
+// "old" and "new" patterns, and we may end up over- or under-escaping things,
+// or in an escaping loop, if things aren't properly ordered.
 add("&lt;", "<");
 add("&gt;", ">");
 add("&amp;", "&"); // &amp; must be after all replacements involving &s.
+
+function replaceAll(string, old, replacement) {
+    return string.split(old).join(replacement);
+}
 
 /**
  * Performs any escaping, unescaping, or substituting required to make the text
@@ -25,7 +33,7 @@ add("&amp;", "&"); // &amp; must be after all replacements involving &s.
 var slackToMatrix = function(string) {
     for (var i = 0; i < substitutionPairs.length; ++i) {
         var pair = substitutionPairs[i];
-        string = string.replace(pair.slack, pair.matrix);
+        string = replaceAll(string, pair.slack, pair.matrix);
     }
     return string;
 };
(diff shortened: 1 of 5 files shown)


lib/substitutions.js
```diff
@@ -0,0 +1,50 @@
...
+
+function add(slack, matrix) {
+    substitutionPairs.push({
+        slack: slack,
+        matrix: matrix
+    });
+}
+
+// slack -> matrix substitutions are performed top -> bottom
+// matrix -> slack substitutions are performed bottom -> top
```
Kegsay: Why?

lib/substitutions.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Performs any escaping, unescaping, or substituting required to make the text
+ * of a Slack message appear like the text of a Matrix message.
+ *
+ * @param {string} string the text, in Slack's format.
+ */
+var slackToMatrix = function(string) {
+    for (var i = 0; i < substitutionPairs.length; ++i) {
+        var pair = substitutionPairs[i];
+        string = string.replace(pair.slack, pair.matrix);
```
Kegsay: Ah yes, `String.replace` is special. This will only actually replace the *first occurrence* of `pair.slack` - You can get around this by specifying a global regex e.g. `string.replace(/foo/g, "bar")` or if you're regex averse.. well.. keep applying it until there are no diffs? :D

lib/substitutions.js
```diff
@@ -0,0 +1,50 @@
...
+    for (var i = substitutionPairs.length - 1; i >= 0; --i) {
+        var pair = substitutionPairs[i];
+        string = string.replace(pair.matrix, pair.slack);
+    }
+    return string;
+};
+
+module.exports = {
+    "matrixToSlack": matrixToSlack,
+    "slackToMatrix": slackToMatrix
```
Kegsay: You don't need the `"` here for the record, provided that the characters aren't funky (e.g. spaces, etc)

lib/slack-hook-handler.js
```diff
@@ -34,7 +36,7 @@ SlackHookHandler.prototype.handle = function(params) {
         if (this.rooms.knowsSlackChannel(params["channel_id"])) {
             var roomID = this.rooms.matrixRoomID(params["channel_id"]);
```
Kegsay: If you're referencing properties literally like this you can just do `params.channel_id` - both forms are valid but because you *have* to do computed property keys in `obj[keyVar]` format, there's a nice(ish) distinction between the two (well-defined structs use `foo.bar` whereas lookup tables use `foo[barVar]`

Kegsay:
LGTM aside from comments about single replacements.

You'll probably find that the current architecture for `substitutions` will get unruly *fast* (e.g. let's add HTML messages, better pass the whole `event` Object rather than just a `string`... oh now let's replace the `user_id` with their display name, guess I need some context on the `Room` as well...) which may make it a lot less functional than you have it currently (if you start needing to do HTTP hits to cache display names, etc). Just something to think about wrt testing.

lib/substitutions.js
illicitonion: Expanded comment.

lib/substitutions.js
illicitonion: Done.

And yes, I am indeed regex averse. Death to regex.

lib/substitutions.js
illicitonion: Done

illicitonion:
Definitely worth bearing in mind :)

PR #6, opened 2015-09-22
Support images matrix -> slack

illicitonion:
But not yet slack -> matrix

[Review on Reviewable](https://reviewable.io/reviews/matrix-org/matrix-appservice-slack/6)

### DIFF
--- a/app.js
+++ b/app.js
@@ -71,7 +71,7 @@ new Cli({
     },
     run: function(port, config) {
         var rooms = new Rooms(config);
-        var matrixHandler = new MatrixHandler(rooms, requestLib);
+        var matrixHandler = new MatrixHandler(config, rooms, requestLib);
         bridge = new Bridge({
             homeserverUrl: config.homeserver.url,
             domain: config.homeserver.server_name,
--- a/lib/matrix-handler.js
+++ b/lib/matrix-handler.js
@@ -7,7 +7,8 @@ var substitutions = require("./substitutions");
  * @param {Rooms} rooms mapping of all known slack channels to matrix rooms.
  * @param {request} requestLib request library, for sending HTTP requests.
  */
-function MatrixHandler(rooms, requestLib) {
+function MatrixHandler(config, rooms, requestLib) {
+    this.config = config;
     this.rooms = rooms;
     this.requestLib = requestLib;
 }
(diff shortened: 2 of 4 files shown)


Kegsay:
LGTM other than comments

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[lib/matrix-handler.js, line 8 \[r1\]](https://reviewable.io:443/reviews/matrix-org/matrix-appservice-slack/6#-Jzozde6zRqFv7nEb9Z3)** ([raw file](https://github.com/matrix-org/matrix-appservice-slack/blob/60b4e5d12b9cdc0bea36cb6ab9390052053bfc9c/lib/matrix-handler.js#L8)):</sup>
Missing JSDoc for new parameter `config`

---

<sup>**[lib/substitutions.js, line 58 \[r1\]](https://reviewable.io:443/reviews/matrix-org/matrix-appservice-slack/6#-Jzozu96yREFEYV7OypA)** ([raw file](https://github.com/matrix-org/matrix-appservice-slack/blob/60b4e5d12b9cdc0bea36cb6ab9390052053bfc9c/lib/substitutions.js#L58)):</sup>
You should default to `===` rather than `==` unless you know what you're doing (read: only for lazy null/undefined checking). 99% of the time, using `==` will be a subtle source of bugs, *especially* when doing `0` checks.

---

<sup>**[tests/matrix-handler.spec.js, line 20 \[r1\]](https://reviewable.io:443/reviews/matrix-org/matrix-appservice-slack/6#-Jzp-RTrBw5Kna0icjbU)** ([raw file](https://github.com/matrix-org/matrix-appservice-slack/blob/60b4e5d12b9cdc0bea36cb6ab9390052053bfc9c/tests/matrix-handler.spec.js#L20)):</sup>
The JS idiom here would be:
```javascript
args.keys().forEach(function(k) {
  event.content[k] = args[k];
}
```

The `in` operator is (mostly) evil, and is a great source of bugs (yum yum!)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/matrix-org/matrix-appservice-slack/6)

illicitonion:
Done

PR #7, opened 2015-09-22
Require auth tokens on slack webhook requests

illicitonion:
[Review on Reviewable](https://reviewable.io/reviews/matrix-org/matrix-appservice-slack/7)

### DIFF
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -12,5 +12,6 @@ tls:
 
 rooms:
   - slack_channel_id: "QWERTYUIOP"
+    slack_api_token: "ABC123"
     matrix_room_id: "!ZXC987:matrix.org"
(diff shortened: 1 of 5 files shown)


config/config.sample.yaml
```diff
@@ -12,5 +12,6 @@ tls:
 
 rooms:
   - slack_channel_id: "QWERTYUIOP"
+    slack_api_token: "ABC123"
```
Kegsay: Are the tokens really scoped per room?!

Kegsay:
LGTM other than comments.

config/config.sample.yaml
illicitonion: They're scoped per hook, and we're currently doing hook per room. bleh.

PR #8, opened 2015-09-22
Basic slack -> emoji support

illicitonion:
Translates slack :foo: into canonical emoji unicode


lib/substitutions.js
```diff
@@ -20,6 +22,22 @@ add("&lt;", "<");
...
+        var e = emoji[i];
+        for (var j = 0; j < e.short_names.length; ++j) {
+            if (e["unified"].indexOf("-") < 0) {
+                var char = String.fromCodePoint(parseInt(e["unified"], 16));
+                index[":" + e.short_names[j] + ":"] = char;
+            }
+        }
+    }
+    return index;
+})();
```
Kegsay: Why are we doing this instead of just assigning directly to `emojiIndex`? Also, it needlessly obfuscates the code making this a named function `indexEmoji` which we never actually use anywhere.

lib/substitutions.js
```diff
@@ -35,6 +53,11 @@ var slackToMatrix = function(string) {
         var pair = substitutionPairs[i];
         string = replaceAll(string, pair.slack, pair.matrix);
     }
+    if (/:[^: ]*:/.test(string)) {
```
Kegsay: maybe replace `[^: ]` with `[^:\s]` to match any whitespace other than spaces?

tests/slack-hook-handler.spec.js
```diff
@@ -55,6 +55,18 @@ describe("SlackHookHandler.handle", function() {
...
+        it("unicode-ifies emoji", function() {
+            intent = createSpyObj("intent", ["sendText"]);
+            handler.handle(makeRequest(
+                    "Lucille",
+                    "nudge nudge:wink: :wink::rugby_football:wink :wink wink:"
+                )
+            );
+            expect(intent.sendText).toHaveBeenCalledWith(
+                "!room:host",
+                "nudge nudge😉 😉🏉wink :wink wink:"
```
Kegsay: It is *extremely* difficult to tell that this should be the expected output. It would be nicer to have multiple tests to assert that it:
 - Can do multiple `:foo:` on a single line
 - Doesn't do `:fo o:` (whitespace). Doesn't do ones with additional `:`s e.g. `:fo:o:`.
 - Gracefully handles unknown strings like `:nomnomnom:` by leaving them alone.

Kegsay:
LGTM other than comments.

lib/substitutions.js
illicitonion: Avoiding polluting the file namespace with random stuff.

In particular, I got complaints from the linter that "i" was already defined in file scope when I didn't have this hidden away in a function.

I agree that the function doesn't need a name, though... :)

tests/slack-hook-handler.spec.js
illicitonion: Done

lib/substitutions.js
illicitonion: Done

tests/slack-hook-handler.spec.js
```diff
@@ -55,6 +55,31 @@ describe("SlackHookHandler.handle", function() {
...
+            expect(intent.sendText).toHaveBeenCalledWith("!room:host", "😉😉 🏉");
+        });
+        it("doesn't unicode-ifies emoji with whitespace", function() {
+            intent = createSpyObj("intent", ["sendText"]);
+            handler.handle(makeRequest("Lucille", ":win k:"));
+            expect(intent.sendText).toHaveBeenCalledWith("!room:host", ":win k:");
+        });
+        it("doesn't unicode-ifies improperly formed emoji", function() {
+            intent = createSpyObj("intent", ["sendText"]);
+            handler.handle(makeRequest("Lucille", ":wink:k:"));
```
Kegsay: This gives the parser an easy ride since it can match in one sweep `:wink:`, we should also do `:k:wink:` to make sure that it discards that first `:` as a potential match.

illicitonion: Done

Kegsay:
LGTM

PR #9, opened 2015-09-23
Support sending images slack -> matrix

illicitonion:
Only supported if a slack_master_token is specified, which is a token
with permission to view history of each room being bridged.

### DIFF
--- a/lib/slack-hook-handler.js
+++ b/lib/slack-hook-handler.js
@@ -97,7 +97,12 @@ SlackHookHandler.prototype.lookupAndSendNonTextMessage =
                 return;
             }
             if (resp.messages.length != 1) {
-                // Just laziness
+                // Just laziness.
+                // If we get unlucky and two messages were sent at exactly the
+                // same microsecond, we could parse them all, filter by user,
+                // filter by whether they have attachments, and such, and pick
+                // the right message. But this is unlikely, and I'm lazy, so
+                // we'll just drop the message...
                 console.log("Really unlucky, got multiple messages at same" +
                     " microsecond, dropping:" + body);
                 return;
--- a/lib/slack-hook-handler.js
+++ b/lib/slack-hook-handler.js
@@ -31,16 +31,21 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
  * Attempts to make the message as native-matrix feeling as it can.
  *
  * @param {Object} params HTTP body of the webhook request, as a JSON-parsed dictionary.
+ * @param {string} params.channel_id Slack channel ID receiving the message.
+ * @param {string} params.channel_name Slack channel name receiving the message.
+ * @param {string} params.user_id Slack user ID of user sending the message.
+ * @param {string} params.user_name Slack user name of the user sending the message.
+ * @param {?string} params.text Text contents of the message, if a text message.
+ * @param {string} timestamp Timestamp when message was received, in seconds
+ *     formatted as a float.
  */
 SlackHookHandler.prototype.handle = function(params) {
     if (params.user_id === "USLACKBOT") {
         return;
     }
     if (!this.rooms.knowsSlackChannel(params.channel_id)) {
-        console.log(
-            "Ignoring message for slack channel " +
-            "with unknown matrix ID: " + params.channel_id +
-            " (" + params.channel_name + ")"
+        console.log("Ignoring message for slack channel with unknown matrix ID: %s (%s)",
+            params.channel_id, params.channel_name
         );
         return;
     }
@@ -50,6 +55,8 @@ SlackHookHandler.prototype.handle = function(params) {
         intent.sendText(roomID, substitutions.slackToMatrix(params.text));
         return;
     }
+    // Attachments are never sent to a webhook, so we need to actually request
+    // the message history of the room to get information about attachments.
     this.lookupAndSendNonTextMessage(params.channel_id, params.timestamp, intent, roomID);
     return;
 };
@@ -68,6 +75,8 @@ SlackHookHandler.prototype.lookupAndSendNonTextMessage =
     if (!this.config.slack_master_token) {
         return;
     }
+    // Look up all messages at the exact timestamp we received.
+    // This has microsecond granularity, so should return the message we want.
     var params = {
         channel: channelID,
         latest: timestamp,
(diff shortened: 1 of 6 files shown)


lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
  * @param {Object} params HTTP body of the webhook request, as a JSON-parsed dictionary.
```
Kegsay: You can document properties of an object by adding lines like:
```
@param {string} params.user_id The slack user's ID
```

This would make things clearer in this function I think.

lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
-                intent.sendText(roomID, substitutions.slackToMatrix(params.text));
+    if (params.user_id === "USLACKBOT") {
+        return;
+    }
+    if (!this.rooms.knowsSlackChannel(params.channel_id)) {
+        console.log(
+            "Ignoring message for slack channel " +
+            "with unknown matrix ID: " + params.channel_id +
+            " (" + params.channel_name + ")"
+        );
```
Kegsay: Nit: I think this looks better:
```javascript
console.log(
    "Ignoring message for slack channel with unknown matrix ID: %s (%s)",
    params.channel_id, params.channel_name
);
```

lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
+    }
+    var params = {
+        channel: channelID,
+        latest: timestamp,
+        oldest: timestamp,
+        inclusive: "1",
+        token: this.config.slack_master_token,
+    };
+    var self = this;
+    this.requestLib.post("https://slack.com/api/channels.history", {form: params},
```
Kegsay: Some comments stating why you're doing this would be great.

lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
+            function(err, _, body) {
+        if (err) {
+            console.log("Error looking up history: %s", err);
+        }
+        else {
+            var resp = JSON.parse(body);
+            if (!resp.ok || !resp.messages) {
```
Kegsay: Sadly this is insufficient to filter out 0-length arrays, so you can NPE on `:95`, you need to add `|| resp.messages.length === 0`

This happens because the check `!resp.messages` just checks to see if there is a truthy value at the property `messages` and an empty array `[]` is truthy.

lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
+                console.log("Could not find history: " + body);
+                return;
+            }
+            if (resp.messages.length != 1) {
+                console.log("Really unlucky, got multiple messages at same" +
+                    " microsecond, dropping:" + body);
```
Kegsay: A bit of explanation here would be good. I understand we're dropping something because `messages.length > 1` (which is only because of the null check earlier, it may be clearer to say `if (resp.messages.length > 1)` instead) but it's not clear that this is due to a race with them having the same `timestamp` value. It would probably also be polite to log the things being dropped, rather than just the raw response and let the poor person looking at the logs to assume/have to look at the code to find out that only the first element is being sent.

lib/slack-hook-handler.js
illicitonion: We then do a length != 1 check below (which, I'll grant you, was missing a return statement, which I've now added); is that sufficient?

lib/slack-hook-handler.js
illicitonion: Done

lib/slack-hook-handler.js
illicitonion: Done

lib/slack-hook-handler.js
illicitonion: Done

lib/slack-hook-handler.js
illicitonion: Done

Not doing the full logging because that would require the parsing which is part of the thing I'm too lazy to do :)

lib/slack-hook-handler.js
```diff
@@ -31,22 +33,120 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
+            console.log("Ignoring non-image attachment: " + body);
+        }
+    });
+};
+
+/**
+ * Converts a slack image attachment to a matrix image event.
+ *
+ * @param {Object} file The slack image attachment file object.
+ * @return {Object} Matrix event content.
```
Kegsay: Docs on `@param` would be cool. Sadly it isn't actually valid JSDoc for the reverse (doccing properties on `@return`), but you can always `@typedef` it e.g. `@return {Foo} A foo` and then have:
```javascript
/**
 * @typedef Foo
 * @property {string} bar A bar
 */
```

Kegsay:
LGTM other than comments. Needs NPE fixing at the very least.

lib/slack-hook-handler.js
Kegsay: Is it ever valid to be continuing if `resp.messages.length === 0`? If not, we should bail early.

illicitonion: Done

lib/slack-hook-handler.js
illicitonion: Done. Rather than documenting the full return value, I just linked to the spec.

lib/slack-hook-handler.js
```diff
@@ -29,24 +31,150 @@ SlackHookHandler.prototype.getIntent = function(slackUser) {
...
+            if (resp.messages.length != 1) {
+                // Just laziness.
+                // If we get unlucky and two messages were sent at exactly the
+                // same microsecond, we could parse them all, filter by user,
+                // filter by whether they have attachments, and such, and pick
+                // the right message. But this is unlikely, and I'm lazy, so
+                // we'll just drop the message...
+                console.log("Really unlucky, got multiple messages at same" +
+                    " microsecond, dropping:" + body);
+                return;
```
Kegsay: Do you intend for this to `return` now?

illicitonion: Yep

PR #10, opened 2015-09-24
Support /me slack -> matrix

illicitonion:


Kegsay:
LGTM with a few non blocking queries:

 * You should get into the habit of defaulting comparisons to `===` because `==` has subtle failure modes. It's not an issue in this PR specifically, but I wouldn't rely on me picking out the failure in a PR as I may miss it. 

 * Why do we suppress the message entirely if we get same microsecond dupes? Can we just take the first and fall through?

PR #11, opened 2015-10-06
Send attachments

illicitonion:


lib/slack-hook-handler.js
```diff
@@ -140,7 +140,16 @@ SlackHookHandler.prototype.lookupAndSendMessage =
             }
         }
         else {
-            console.log("Ignoring non-image attachment: " + body);
+            console.log("Ignoring non-text non-image message: " + body);
```
Kegsay: This is confusing because you're handling text beneath the `else` but saying you're ignoring it without a `return` from this block. It would be much clearer to make this function work for the the types you expect it to and bail early if it isn't a known type.

PR #53, opened 2017-10-01
Replace Riot pills mention to slack mention

### LINKED ISSUE #52: Pills do not highlight Slack users
It's understandable that simply typing a user's display name isn't converted to a Slack-side highlight, as that would likely result in false positives. However, the recent "Pill" behavior introduced in Riot offers a clear signal that a user was intentionally denoted. It would be nice to support this.

Currently, in order to ping a Slack user, one has to use an @ prefix to their name - and this does not work together with tab completion.

erdnaxeli:
Fix #52

Actually if you write `"someone <a href="…">someone</a>`, the first occurrence of "someone" will get converted to "@someone" instead of the 2d, but it does not seem to much annoying to add the complexity to handle it.

### DIFF
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -188,6 +188,10 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
+Main.prototype.getUsernamePrefix = function() {
+    return this._config.username_prefix;
+};
+
 // Returns a Promise of a SlackGhost
 Main.prototype.getGhostForSlackMessage = function(message) {
     // Slack ghost IDs need to be constructed from user IDs, not usernames,
--- a/lib/substitutions.js
+++ b/lib/substitutions.js
@@ -115,6 +115,20 @@ var matrixToSlack = function(event, main) {
         if (event.content.msgtype == "m.emote") {
             string = "_" + string + "_";
         }
+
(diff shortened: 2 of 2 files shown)


erdnaxeli:
@lukebarnard1 does this regex looks ok to you?

lukebarnard1:
I think this is fine for the most part. The only weirdness is when a client decides to send something in the text of the `<a>` tag that isn't the display name of the user. There's nothing in the protocol that enforces this - Riot deals with this by replacing the text with the known display name of the user.

If you really care, you could use the `userId` (in the matrix.to `href`) to look up the display name of the user, and then user that to form the "@DisplayName".

Cadair:
I have been testing this, and it appears that it's possible for someone on the slack side to have a display name and a username that don't match. The more robust way of doing this to my mind would be to extract the slack user id string from the localpart of the mxid and convert that into the username and send that in the "@someone" to the slack side. I don't know how practical this is.

Cadair:
I have updated this a little in erdnaxeli/matrix-appservice-slack#3 I think this also semi-resolved @lukebarnard1's concern as now only the body of the link and the plain text version of the mention have to match each other and not the actual username of the matrix user.

Cadair:
This can be closed as it is replaced by #96

Issue #29, opened 2016-09-15
Work out how we can identify ourselves and not ignore other webooks


leonerd:
The gitter relaybot acts as a real bot user with a real user ID, so it can recognise reflections via gitter of its own messages, and ignore those on the inbound path.

By comparison, there is no slack user corresponding to our own specific webhook integration. All webhook integrations always appear under the special user ID of "USLACKBOT". This means that the logic to identify our own reflections also rejects every other message from any other webhook integration.

It's possible we can use the `bot_id` field.

leonerd:
This might be possible to solve if we use `bot` OAuth2 scope instead of an incoming webhook. When we create that bot scope Slack tells us what the bot ID is.

Half-Shot:
We do this now.

tadzik:
We do?

It still looks hardcoded [here](https://github.com/matrix-org/matrix-appservice-slack/blob/develop/src/SlackHookHandler.ts#L273)

PR #64, opened 2017-12-18
Allow syncing Bot messages from Slack to Matrix

### LINKED ISSUE #42: Messages from other apps in Slack (that use attachments) are silently ignored
Hey,
just noticed that messages from apps like the RSS integration are silently ignored. This might be related to #37 as both are not ordinary messages.
This can cause confusing conversations when other team members refer to something that was posted by another app ;)

MTRNord:
This is untested! (I have no HS access to test but it is according to the slack Spec https://api.slack.com/events/message/bot_message )

This should fix #42 but as said I am not able to test it myself

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -189,6 +189,10 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
         var text = substitutions.slackToMatrix(message.text);
         return ghost.sendText(roomID, text);
     }
+    else if (subtype === "bot_message") {
+        var text = substitutions.slackToMatrix(message.text);
+        return ghost.sendText(roomID, text);
+    }
     else if (subtype === "me_message") {
         var message = {
             msgtype: "m.emote",


erdnaxeli:
I try this this week-end.

erdnaxeli:
Okay, it's more complex than that, see #29. We need to move from outgoing-webhook to a real bot user.

MTRNord:
@erdnaxeli hm ok :/ the doc/spec meant it should work also with webhooks :( but ok I will close this to not cause confusion

MTRNord:
grafik

(Event API should be Webhooks)

MTRNord:
nvm apperantly Webhooks ARE NOT Events Api :/

erdnaxeli:
This is not the problem. The event is correctly received, but the code ignore it (https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/SlackHookHandler.js#L182).

MTRNord:
Oh Hm well ok i will Take a look at making that check better.

MTRNord:
Looking at the Code we already can get our bot ID when registering the webhook/getting the OAuth2 access_token. This is documented in: https://api.slack.com/bot-users Under the "Retrieving your bot user token" Section. In there is a "bot_user_id" field which should contain the ID of our bot and should be useable. I will see if I can implement all the stuff needed to "fix"/change the if you refered too.

PR #66, opened 2018-01-13
add support for slack Events api

perissology:
This PR updates the bridge to use slacks events api & a bot user. This allows bridging of as many channels as you would like with only 1 slack integration per slack team/workspace.

This is fully backwards compatible with the incoming/outgoing webhooks.

Also improvea slack -> matrix file uploads

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,21 @@ The bridge itself should now be running.
 
 To actually use it, you will need to configure some linked channels.
 
-Slack Bot Setup
----------------
+Provisioning
+------------
+
+This bridge allows linking together pairs of Matrix rooms and Slack channels,
+relaying messages said by people in one side into the other. To create a link
+first the individual Matrix room and Slack channel need to be created, and then
+a command needs to be issued in the administration console room to add the link
+to the bridge's database.
+
+There are 2 ways to bridge a room. The recommended way uses the newer slack events api
+and bot users. This allows you to link as many channels as you would like with only
+1 slack integration. The legacy way uses incoming/outgoing webhooks, and requires
+2 slack integrations per channel to be bridged.
+
+### Recommended 
 
 1. add a custom app to your slack team/workspace by visiting https://api.slack.com/apps
    and clicking on `Create New App`.
@@ -113,23 +126,39 @@ Slack Bot Setup
 5. Click on `Install App` and `Install App to Workspace`. Note the `Bot User OAuth Access Token`
    as you will need it whenever you link a room.
    
-6. You will need to invite the bot user you created above to each slack channel you would like
-   to bridge.
+6. For each channel you would like to bridge, perform the following steps:
+
+   1. Create a Matrix room in the usual manner for your client. Take a note of its
+      Matrix room ID - it will look something like `!aBcDeF:example.com`.
+      
+   2. invite the bot user to the slack channel you would like to bridge.
    
-   ```
-   /invite @bot-user-name
+       ```
+       /invite @bot-user-name
 
-   ``` 
+       ``` 
+       
+       You will also need to determine the "channel ID" that Slack uses to identify
+       the channel. Unfortunately, it is not easily obtained from the Slack UI. The
+       easiest way to do this is to send a message from Slack to the bridge; the
+       bridge will log the channel ID as part of the unrecognised message output.
+       You can then take note of the `channel_id` field.
+    
+   2. Issue a ``link`` command in the administration control room with these
+      collected values as arguments:
+      
+     ```
+     link --channel_id CHANNELID --room !the-matrix:room.id --slack_token xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxxx
+     ```
+  
+     These arguments can be shortened to single-letter forms:
+  
+     ```
+     link -I CHANNELID -R !the-matrix:room.id -t xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxxx
+     ```
 
 
-Provisioning
-------------
-
-This bridge allows linking together pairs of Matrix rooms and Slack channels,
-relaying messages said by people in one side into the other. To create a link
-first the individual Matrix room and Slack channel need to be created, and then
-a command needs to be issued in the administration console room to add the link
-to the bridge's database.
+### Legacy
 
 1. Create a Matrix room in the usual manner for your client. Take a note of its
    Matrix room ID - it will look something like `!aBcDeF:example.com`.
@@ -154,13 +183,13 @@ to the bridge's database.
    collected values as arguments:
 
    ```
-   link --channel CHANNELID --room !the-matrix:room.id --token THETOKEN --webhook_uri http://the.webhook/uri
+   link --channel_id CHANNELID --room !the-matrix:room.id --webhook_url https://hooks.slack.com/services/ABC/DEF/123
    ```
 
    These arguments can be shortened to single-letter forms:
 
    ```
-   link -c CHANNELID -r !the-matrix:room.id -t THETOKEN -u http://the.webhook/uri
+   link -I CHANNELID -R !the-matrix:room.id -u https://hooks.slack.com/services/ABC/DEF/123
    ```
 
 See also https://github.com/matrix-org/matrix-appservice-bridge/blob/master/HOWTO.md for the general theory of all this :)
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ and bot users. This allows you to link as many channels as you would like with o
 
 ### Recommended 
 
-1. add a custom app to your slack team/workspace by visiting https://api.slack.com/apps
+1. Add a custom app to your slack team/workspace by visiting https://api.slack.com/apps
    and clicking on `Create New App`.
    
 2. Name the app & select the team/workspace this app will belong to.
@@ -129,10 +129,11 @@ and bot users. This allows you to link as many channels as you would like with o
 
    - files:write:user
    
-   Note: media uploaded to matrix is currently not permissioned, and anyone with the link
-   can access the file. In order to make slack files visible to matrix users, this bridge
-   will set make the slack file visible to anyone with the url (including files in private channels).
-   The current behavior in slack is that files are only accessible to authenticated users.
+   Note: any media uploaded to matrix is currently accessible by anyone who knows the url.
+   In order to make slack files visible to matrix users, this bridge will make slack files
+   visible to anyone with the url (including files in private channels). This is different
+   then the current behavior in slack, which only allows authenticated access to media
+   posted in private channels.
  
 6. Click on `Install App` and `Install App to Workspace`. Note the access tokens show.
    You will need the `Bot User OAuth Access Token` and if you want to bridge files, the
@@ -151,10 +152,7 @@ and bot users. This allows you to link as many channels as you would like with o
        ``` 
        
        You will also need to determine the "channel ID" that Slack uses to identify
-       the channel. Unfortunately, it is not easily obtained from the Slack UI. The
-       easiest way to do this is to send a message from Slack to the bridge; the
-       bridge will log the channel ID as part of the unrecognised message output.
-       You can then take note of the `channel_id` field.
+       the channel, which can be found in the url `https://XXX.slack.com/messages/<channel id>/`.
     
    2. Issue a ``link`` command in the administration control room with these
       collected values as arguments:
--- a/README.md
+++ b/README.md
@@ -122,11 +122,23 @@ and bot users. This allows you to link as many channels as you would like with o
      
        - team_domain_change
        - message.channels
+       - message.groups (if you want to bridge private channels)
        
-5. Click on `Install App` and `Install App to Workspace`. Note the `Bot User OAuth Access Token`
-   as you will need it whenever you link a room.
+5. Skip this step if you do not want to bridge files.
+   Click on `OAuth & Permissions` and add the following scopes:
+
+   - files:write:user
    
-6. For each channel you would like to bridge, perform the following steps:
+   Note: media uploaded to matrix is currently not permissioned, and anyone with the link
+   can access the file. In order to make slack files visible to matrix users, this bridge
+   will set make the slack file visible to anyone with the url (including files in private channels).
+   The current behavior in slack is that files are only accessible to authenticated users.
+ 
+6. Click on `Install App` and `Install App to Workspace`. Note the access tokens show.
+   You will need the `Bot User OAuth Access Token` and if you want to bridge files, the
+   `OAuth Access Token` whenever you link a room.
+   
+7. For each channel you would like to bridge, perform the following steps:
 
    1. Create a Matrix room in the usual manner for your client. Take a note of its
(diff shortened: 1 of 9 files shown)


lib/BridgedRoom.js
```diff
@@ -300,4 +391,68 @@ BridgedRoom.prototype.getMatrixATime = function() {
...
+    if (!this._slack_bot_token) return Promise.resolve();
+
+    return rp({
+        uri: 'https://slack.com/api/team.info',
+        qs: {
+            token: this._slack_bot_token
+        },
+        json: true,
+    }).then((response) => {
+        if (!response.team) return;
```
erdnaxeli: Why not `Promise.resolve()`?

lib/BridgedRoom.js
```diff
@@ -300,4 +391,68 @@ BridgedRoom.prototype.getMatrixATime = function() {
...
+            this._slack_team_domain = response.team.domain;
+            this._dirty = true;
+        }
+
+        if (!this._slack_team_id !== response.team.id) {
+            this._slack_team_id = response.team.id;
+            this._dirty = true;
+        }
+
+        return;
```
erdnaxeli: Same question here.

lib/BridgedRoom.js
```diff
@@ -300,4 +391,68 @@ BridgedRoom.prototype.getMatrixATime = function() {
...
+        });
+    }).then((response) => {
+        if (!response.user && response.user.profile) return Promise.resolve();
+
+        if (!this._slack_bot_id !== response.user.profile.bot_id) {
+            this._slack_bot_id = response.user.profile.bot_id;
+            this._dirty = true;
+        }
+
+        return;
```
erdnaxeli: And here.

lib/Main.js
```diff
@@ -193,6 +194,7 @@ Main.prototype.getGhostForSlackMessage = function(message) {
     // Slack ghost IDs need to be constructed from user IDs, not usernames,
     //   because users can change their names
 
+    // TODO it the team_domain is changed, we will recreate all users
```
erdnaxeli: s/it/if/

lib/Main.js
```diff
@@ -381,6 +397,7 @@ Main.prototype.listRoomsFor = function(intent, state) {
 };
 
 Main.prototype.onMatrixEvent = function(ev) {
+    console.log("Received matrix event:", JSON.stringify(ev));
```
erdnaxeli: I think this will be too much verbose. We should use some debug mode and not always enable it.

lib/Main.js
```diff
@@ -521,6 +539,32 @@ Main.prototype.actionLink = function(opts) {
...
+        room.updateSlackUserToken(opts.slack_user_token);
+    }
+
+    if (opts.slack_bot_token) {
+        room.updateSlackBotToken(opts.slack_bot_token);
+        return Promise.all([ room.lookupAndSetTeamInfo(), room.lookupAndSetUserInfo() ])
+            .then(() => {
+                if (isNew) {
+                    this.addBridgedRoom(room);
+                }
```
erdnaxeli: You should write it on one line, as you do on line 566.

lib/Main.js
```diff
@@ -521,6 +539,32 @@ Main.prototype.actionLink = function(opts) {
...
+
+                if (room.isDirty()) {
+                    this.putRoomToStore(room);
+                }
```
erdnaxeli: Same here.

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+
+        // respond to event url challenges
+        if (params.type === 'url_verification') {
+            response.writeHead(200, {"Content-Type": "application/json"});
+            response.write(JSON.stringify({challenge: params.challenge}));
+            response.end();
+            return;
+        }
+
+        // var room = main.getRoomBySlackChannelId(params.event.channel);
```
erdnaxeli: dead code

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+ * Handles a slack event request.
+ *
+ * @param {Object} params HTTP body of the event request, as a JSON-parsed dictionary.
+ * @param {string} params.team_id The unique identifier for the workspace/team where this event occurred.
+ * @param {Object} params.event Slack event object
+ * @param {string} params.event.type Slack event type
+ * @param {string} params.type type of callback we are receiving. typically event_callback
+ *     or url_verification.
+ // * @param {string} timestamp Timestamp when message was received, in seconds
+ // *     formatted as a float.
```
erdnaxeli: You can remove it.

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+                break;
+            case 'file_comment_added':
+                result = Promise.resolve();
+                break;
+            default:
+                result = Promise.reject(new UnknownEvent());
+        }
+
+        result
+            .then(() => endTimer({outcome: "success"}))
```
erdnaxeli: `result.then` on one line, no?

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+            );
+
+        response.writeHead(200, {"Content-Type": "application/json"});
+        response.end();
+    } catch (e) {
+        console.log("Oops - SlackEventHandler failed:", e);
+
+        // return 200 so slack doesn't keep sending the event
+        response.writeHead(200, {"Content-Type": "text/plain"});
+        response.end();
```
erdnaxeli: You should use `finally` to avoid duplicated code.

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * Attempts to handle the `message` event.
+ *
+ * Sends a message to Matrix if it understands enough of the message to do so.
+ * Attempts to make the message as native-matrix feeling as it can.
+ *
+ * @param {Object} params The event request emitted.
+ * @param {string} params.event.user Slack user ID of user sending the message.
+ // * @param {string} params.user_name Slack user name of the user sending the message.
```
erdnaxeli: If it's not needed anymore, just remove it.

lib/SlackEventHandler.js
```diff
@@ -0,0 +1,222 @@
...
+ * @param {?string} params.event.text Text contents of the message, if a text message.
+ * @param {string} params.event.channel The slack channel id
+ * @param {string} params.event.ts The unique (per-channel) timestamp
+ */
+SlackEventHandler.prototype.handleMessageEvent = function (params) {
+    var room = this._main.getRoomBySlackChannelId(params.event.channel);
+    if (!room) throw new UnknownChannel(params.event.channel);
+
+    if (params.event.subtype === 'bot_message' && params.event.bot_id === room.getSlackBotId()) {
```
erdnaxeli: If there is not `slack_bot_token`, `lookupAndSetUserInfo` will never set `slack_bot_id`, so this condition is always true, and I'm afraid we have a loop.

The condition should be something like:
```Javascript
params.event.subtype === 'bot_message' && (!room.getSlackBotToken || params.event.bot_id === room.getSlackBotId())
```

lib/SlackGhost.js
```diff
@@ -57,20 +57,50 @@ SlackGhost.prototype.update = function(message, room) {
...
-    });
+    var getDisplayName = (!display_name) ?
+        this.lookupUserInfo(message.user_id, room.getAccessToken())
+            .then(user => {
+                if (user && user.profile) {
+                    return user.profile.display_name || user.profile.real_name;
+                }
+                return undefined;
+            }) :
+        Promise.resolve(display_name);
```
erdnaxeli: It not a one liner anymore, please use
```Javascript
var getDisplayName;
if (!display_name) {
    …
}
else {
    result = Promise.resolve(display_name);
}
```

lib/SlackGhost.js
```diff
@@ -57,20 +57,50 @@ SlackGhost.prototype.update = function(message, room) {
...
+
+    return getDisplayName.then(display_name => {
+        if (!display_name) return Promise.resolve();
+        if (this._display_name === display_name) return Promise.resolve();
```
erdnaxeli: You can do both with one `if`.

lib/SlackHookHandler.js
```diff
@@ -53,9 +59,15 @@ SlackHookHandler.prototype.startAndListen = function(port, tls_config) {
             });
 
             request.on("end", () => {
-                var params = qs.parse(body);
+                var isEvent = request.headers['content-type'] === 'application/json' && request.method === 'POST';
```
erdnaxeli: If `isEvent` is true, this means it's an event from the Event API right? Could you just add a comment please?

lib/SlackEventHandler.js
erdnaxeli: I will let you answer, but if I understand your code we got here only if we are using the Event API, so `slack_bot_id` _is_ defined.

lib/BridgedRoom.js
perissology: there's no need, since we're already in a promise

lib/SlackEventHandler.js
perissology: yes, the `SlackEventHandler` should only be used if `slack_bot_id` is set. I added your check anyways, as it should save some loops when incorrectly setup

perissology:
Thanks for the 2nd pair of eyes and catching all the things I forgot to cleanup :)

lib/BridgedRoom.js
erdnaxeli: Ok, but your are doing it line 447.

lib/BridgedRoom.js
```diff
@@ -137,18 +187,28 @@ BridgedRoom.prototype.toEntry = function() {
...
 
+        if (this._slack_bot_token) {
+            sendMessageParams.headers = {
+                Authorization: 'Bearer ' + this._slack_bot_token
+            };
+            sendMessageParams.body.as_user = false;
+            sendMessageParams.body.channel = this._slack_channel_id;
+        }
+
         sendMessageParams.body.username = user.getDisplaynameForRoom(message.room_id);
```
erdnaxeli: Actually I am wondering if this will still work with a bot user. See here the part about "Authoring messages": https://api.slack.com/changelog/2017-09-the-one-about-usernames#changing.

If I am correct, a bot user cannot change the username anymore :(

Cadair: I did some reading and this page seemed to suggest you could: https://api.slack.com/methods/chat.postMessage (I have no idea just providing links 😆 )

Cadair: Hmm, it looks like maybe they are phasing out that support, which would be disasterous for this AS, both in webhook and bot mode.

perissology: The username & avatar changes definitely work as of now, not sure what will happen in the future.

lib/BridgedRoom.js
perissology: oops. removed.

lib/Main.js
```diff
@@ -422,8 +422,11 @@ Main.prototype.onMatrixEvent = function(ev) {
         // A membership event about myself
         var membership = ev.content.membership;
         if (membership === "invite") {
-            // Automatically accept all invitations
-            this.getBotIntent().join(ev.room_id);
+            // Automatically accept all invitations, but wait 30 seconds before doing it.
+            // this was throwing an error when the bot was invited to a room by a user on
+            // a different homeserver ({"errcode":"M_FORBIDDEN","error":"You are not invited to this room."})
+            Promise.delay(30 * 1000).then(() => this.getBotIntent().join(ev.room_id));
+
```
erdnaxeli: Empty line.

And this feels like a bug in Synapse :s. I wonder if it can be reproduced with a normal client or if it is related to the appservice API.

erdnaxeli:
The in / out webhooks still work, that's good. I test to link a room with event API tomorrow.

erdnaxeli:
When issuing a link command, I got this message: `@erdnaxeli:cervoi.se: Cannot link - TypeError: Cannot read property 'profile' of undefined`.

lib/BridgedRoom.js
```diff
@@ -300,4 +391,64 @@ BridgedRoom.prototype.getMatrixATime = function() {
...
+            qs: {
+                token: this._slack_bot_token,
+                user: response.user_id,
+            },
+            json: true,
+        });
+    }).then((response) => {
+        if (!response.user && response.user.profile) return;
+
+        if (!this._slack_bot_id !== response.user.profile.bot_id) {
```
erdnaxeli: The error came from here, because response his:
```
{ ok: false,
  error: 'missing_scope',
  needed: 'users:read',
  provided: 'identify,bot,channels:history,groups:history,team:read,files:write:user' }
```

erdnaxeli: I added "users:read" in oauth permission, and "user_change" in events, but now when I add the link, I got the message "Row is now pending-name" but the room does not show up in rooms list and link is not working.

erdnaxeli: The error came from here, because response his:
```
{ ok: false,
  error: 'missing_scope',
  needed: 'users:read',
  provided: 'identify,bot,channels:history,groups:history,team:read,files:write:user' }
```

I added "users:read" in oauth permission, and "user_change" in events, but now when I add the link, I got the message "Row is now pending-name" but the room does not show up in rooms list and link is not working.

perissology: are you providing both the bot & user tokens? `--slack_bot_token xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxxx --slack_user_token xoxp-xxxxxxxx-xxxxxxxxx-xxxxxxxx-xxxxxxxx`

perissology: and did you have the `files:write:user` permission. I don't think you would need this, but it would help me debug if I knew how your's was set

erdnaxeli:
I provided both, and I added the file write permission.

Cadair:
Any progress on this one? I have a slack team that's run out of integrations, and I could really do with this!

README.md
```diff
@@ -100,6 +99,84 @@ first the individual Matrix room and Slack channel need to be created, and then
...
+1. Add a custom app to your slack team/workspace by visiting https://api.slack.com/apps
+   and clicking on `Create New App`.
+   
+2. Name the app & select the team/workspace this app will belong to.
+
+3. Click on `bot users` and add a new bot user. We will use this account to bridge the
+   the rooms.
+   
+4. Click on `Event Subscriptions` and enable them. At this point, the bridge needs to be
+   started as slack will do some verification of the request rul. The request url should be
```
PombeirP: typo `rul`=`url`

ara4n:
@Half-Shot when you rejoin us next week please can you put this at the top of your todo list?

Half-Shot:
Sure.

Cadair:
Once this is in I can update my branches and send a couple more PRs with support for gitter-as style edits etc.

Half-Shot:
Ooh, yes that would be exciting @Cadair

Half-Shot:
Thanks @perissology for the hard work in doing this! I've continued the PR on #89.

PR #80, opened 2018-07-13
Fix listRoomsFor and disable invite fetching

Half-Shot:


anoadragon453:
Code should be removed rather than commented out imo

Half-Shot:
Removed code is for functionality we'd like to remove. I don't really want to remove the functionality, just disable it because it's broken. If we eventually get a *sensible* endpoint in this spec then it can be fixed.

lib/Main.js
```diff
@@ -354,29 +354,17 @@ Main.prototype.drainAndLeaveMatrixRoom = function(roomId) {
...
-                '},' +
-                '"presence":{"types":[]},' +
-                '"account_data":{"types":[]}}',
-        }
+    if (state && state !== "join") { 
+        return Promise.reject("listRoomsFor only supports 'join'");
+    }
+    
+    // This endpoint sucks because we can only get joined rooms.
+    // Previously we filtered a /sync but obviously that was broken too.
```
anoadragon453: Would probably keep a TODO here.

PR #82, opened 2018-07-26
Add option for media_url

Half-Shot:
Some homeservers only allow their AS to connect to them over a private network, but we want a public url for media.

### DIFF
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -180,7 +180,9 @@ Main.prototype.putUserToStore = function(user) {
 };
 
 Main.prototype.getUrlForMxc = function(mxc_url) {
-    return this._config.homeserver.url + "/_matrix/media/v1/download/" +
+    const hs = this._config.homeserver;
+    const url = hs.media_url !== undefined ? hs.media_url : hs.url;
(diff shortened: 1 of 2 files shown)


lib/Main.js
```diff
@@ -180,7 +180,9 @@ Main.prototype.putUserToStore = function(user) {
 };
 
 Main.prototype.getUrlForMxc = function(mxc_url) {
-    return this._config.homeserver.url + "/_matrix/media/v1/download/" +
+    const hs = this._config.homeserver;
+    const url = hs.media_url !== undefined ? hs.media_url : hs.url;
```
turt2live: can probably just get away with `hs.media_url || hs.url` or `hs.media_url ? hs.media_url : hs.url` if you're worried about accidental booleans.

lib/Main.js
```diff
@@ -180,7 +180,9 @@ Main.prototype.putUserToStore = function(user) {
...
+    return url + "/_matrix/media/v1/download/" +
```
turt2live: while we're here, s/v1/r0 please

PR #84, opened 2018-08-28
Look in userstore if users.info didn't find a result.

Half-Shot:
Also use pills rather than just names. We now only fail if the user has never spoken on slack and isn't in ``users.info``. We also log why we failed to find the user, because I mistrust the api.

### DIFF
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -191,16 +191,21 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
+Main.prototype.getUserId = function(id, team_domain) {
+    return [
+        "@", this._config.username_prefix, team_domain.toLowerCase(),
+            "_", id.toUpperCase(), ":", this._config.homeserver.server_name
+    ].join("");
+
+}
+
 // Returns a Promise of a SlackGhost
 Main.prototype.getGhostForSlackMessage = function(message) {
     // Slack ghost IDs need to be constructed from user IDs, not usernames,
     //   because users can change their names
 
     // TODO(paul): Steal MatrixIdTemplate from matrix-appservice-gitter
-    var user_id = [
-        "@", this._config.username_prefix, message.team_domain.toLowerCase(),
-            "_", message.user_id.toUpperCase(), ":", this._config.homeserver.server_name
-    ].join("");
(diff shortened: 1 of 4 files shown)


lib/Main.js
```diff
@@ -191,16 +191,21 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
+Main.prototype.getUserId = function(id, team_domain) {
+    return [
+        "@", this._config.username_prefix, team_domain.toLowerCase(),
+            "_", id.toUpperCase(), ":", this._config.homeserver.server_name
```
turt2live: should start considering making this lowercase, although that's a PR for a later time.

lib/Main.js
```diff
@@ -191,16 +191,21 @@ Main.prototype.getBotIntent = function() {
...
+    ].join("");
+
```
turt2live: minor: whitespace

lib/SlackHookHandler.js
```diff
@@ -343,17 +343,42 @@ SlackHookHandler.prototype.replaceUserIdsWithNames = function(message, token) {
...
                 message.text = message.text.replace(/<@(\w+)\|?\w*?>/, response.user.name);
+                message.markdownText = message.markdownText.replace(
+                    /<@(\w+)\|?\w*?>/,
+                    pill
+                );
+                return;
             }
-            else {
-                console.log("users.info returned no result for " + id);
+            console.log("users.info returned no result for " + id+ ". Response:", response);
```
turt2live: add whitespace around `id` please (before the `+`)

lib/SlackHookHandler.js
```diff
@@ -343,17 +343,42 @@ SlackHookHandler.prototype.replaceUserIdsWithNames = function(message, token) {
...
+        }).then((result) => {
+            if (result === undefined) {
+                return;
             }
-            iteration++;
-            }).catch((err) => {
-               console.log("Caught error " + err);
-            });
+            console.log(`${user_id} did ${result.length > 0 ? "not" : ""} an entry`);
+            if(result.length) {
```
turt2live: `if (...` please (consistency)

PR #85, opened 2018-08-28
Pin slackdown to recieve support for triple `

Half-Shot:
We're pinning this because the author doesn't seem all that responsive :/


turt2live:
retroactive approval

Half-Shot:
(sorry)

PR #87, opened 2018-09-12
Allow deep paths to handle incoming hooks

martindale:
This allows the service to handle hooks sent to deep paths, i.e., `/services/matrix/:id` — useful when routing a single public domain to multiple bridges.


Half-Shot:
I think this is good and harmless, @turt2live for a second opinion?

Half-Shot:
I'm going to close this, because our `develop` branch is now Typescript based and has diverged significantly from this PR. Sorry for not getting to this at the time. If the feature is still missing on `develop`, please re-open.

PR #91, opened 2018-09-26
A whole bunch of hacks and improvements

Cadair:
I can spiit this up into multiple bits, but I have tried to keep the commit history tidy. This does:

Now in #92:
* ~Does some html escaping~
* ~**Fixes file bridging because the slack API changed**~
* ~Strips out the language specifiers in triple backticks because they don't render on slack correctly.~

In this PR once #92 is merged:
* Improves pills substitutions to use the new slack mentions
* Add Gitter style edit bridging
* Adds a config option to specify the directory the `user-store.db` and `room-store.db` files go in
* Converts slack snippets to inline code dumps in matrix

### DIFF
--- a/config/slack-config-schema.yaml
+++ b/config/slack-config-schema.yaml
@@ -1,6 +1,8 @@
 type: object
 required: ["homeserver", "slack_hook_port", "inbound_uri_prefix", "bot_username", "username_prefix"]
 properties:
+  dbdir:
+    type: string
   homeserver:
     type: object
     required: ["server_name", "url"]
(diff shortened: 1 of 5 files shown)


Cadair:
Oh also this is a PR to the #89 because I have been using and developing on that for a while.

turt2live:
@Cadair can the changes specific to fixing #89 be split out as a PR to #89 for @Half-Shot to review/incorporate? Everything else pointing towards master would be awesome, and ideally be broken out into (tiny) individual PRs as well, but I won't fight too hard for that.

Cadair:
Ok this now builds on #92 and should be easy enough to split out the commits into their own branches if needed.

PR #94, opened 2018-10-04
Add (winston) logging support

Half-Shot:
This uses the new logging component for matrix-appservice-bridge.

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -4,6 +4,7 @@ var Promise = require('bluebird');
 
 var substitutions = require("./substitutions");
 var rp = require('request-promise');
+const log = require("matrix-appservice-bridge").Logging.get("BridgedRoom");
 
 function BridgedRoom(main, opts) {
     this._main = main;
@@ -163,7 +164,7 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
         return rp(sendMessageParams).then((res) => {
             this._main.incCounter("sent_messages", {side: "remote"});
             if (!res) {
-                console.log("HTTP Error: %s", res);
+                log.error("HTTP Error: %s", res);
             }
         });
     });
@@ -202,7 +203,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
     }
     else if (subtype === "file_share") {
         if (!message.file) {
-            console.log("Ignoring non-text non-image message: " + res);
+            log.info("Ignoring non-text non-image message: " + res);
             return;
         }
         if (message.file._content) {
@@ -227,7 +228,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
         }
     }
     else {
-        console.log("Ignoring message with subtype: " + subtype);
+        log.info("Ignoring message with subtype: " + subtype);
     }
 };
 
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -164,7 +164,7 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
         return rp(sendMessageParams).then((res) => {
             this._main.incCounter("sent_messages", {side: "remote"});
             if (!res) {
-                log.error("HTTP Error: %s", res);
+                log.error("Outgoing message HTTP error: %s", res);
             }
         });
     });
@@ -203,7 +203,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
     }
     else if (subtype === "file_share") {
         if (!message.file) {
-            log.info("Ignoring non-text non-image message: " + res);
+            log.warn("Ignoring non-text non-image message: " + res);
             return;
         }
         if (message.file._content) {
@@ -228,7 +228,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
         }
     }
     else {
-        log.info("Ignoring message with subtype: " + subtype);
+        log.warn("Ignoring message with subtype: " + subtype);
     }
 };
 
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -17,6 +17,7 @@ var OAuth2 = require("./OAuth2");
 var Provisioning = require("./Provisioning");
 
 var randomstring = require("randomstring");
+const log = require("matrix-appservice-bridge").Logging.get("Main");
 
 function Main(config) {
     var self = this;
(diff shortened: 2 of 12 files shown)


lib/Main.js
```diff
@@ -17,6 +17,7 @@ var OAuth2 = require("./OAuth2");
 var Provisioning = require("./Provisioning");
```
turt2live: out of interest, what's the reason for having all the imports broken apart like this?

lib/BridgedRoom.js
```diff
@@ -206,7 +207,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
     }
     else if (subtype === "file_share") {
         if (!message.file) {
-            console.log("Ignoring non-text non-image message: " + res);
+            log.info("Ignoring non-text non-image message: " + res);
```
turt2live: warn? warn is used for other 'Ignoring...' messages later on in the PR

lib/BridgedRoom.js
```diff
@@ -231,7 +232,7 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
         }
     }
     else {
-        console.log("Ignoring message with subtype: " + subtype);
+        log.info("Ignoring message with subtype: " + subtype);
```
turt2live: warn? warn is used for other 'Ignoring...' messages later on in the PR

lib/Main.js
```diff
@@ -409,7 +410,7 @@ Main.prototype.onMatrixEvent = function(ev) {
     if (ev.type === "m.room.member"
         && this._bridge.getBot()._isRemoteUser(ev.state_key)
         && ev.sender !== myUserId) {
-        console.log(`${ev.state_key} got invite for ${ev.room_id} but we can't do DMs, warning room.`);
+        log.debug(`${ev.state_key} got invite for ${ev.room_id} but we can't do DMs, warning room.`);
```
turt2live: warn or info? Seems like something that is a little less debug-y

lib/Main.js
```diff
@@ -419,7 +420,7 @@ Main.prototype.onMatrixEvent = function(ev) {
         }).then(() => {
             return intent.leave(ev.room_id);
         }).catch((err) => {
-            console.warn("Couldn't send warning to user(s) about not supporting PMs", err);
+            log.warn("Couldn't send warning to user(s) about not supporting PMs", err);
```
turt2live: error?

lib/Main.js
```diff
@@ -452,7 +453,7 @@ Main.prototype.onMatrixEvent = function(ev) {
     room.onMatrixMessage(ev).then(
         () => endTimer({outcome: "success"}),
         (e) => {
-            console.log("Failed: ", e);
+            log.error("Failed: ", e);
```
turt2live: While we're here, it might be worth reconsidering the message here. What failed?

lib/Main.js
```diff
@@ -568,7 +569,7 @@ Main.prototype.run = function(port) {
         })
     }).then((entries) => {
         entries.forEach((entry) => {
-            console.log("Ignoring LEGACY room entry in room-store.db", entry);
+            log.error("Ignoring LEGACY room entry in room-store.db", entry);
```
turt2live: warn?

lib/Main.js
```diff
@@ -587,7 +588,7 @@ Main.prototype.run = function(port) {
                 this._stateStorage.trackRoom(entry.matrix_id);
             }
             else {
-                console.log("Ignoring LEGACY room link entry", entry);
+                log.error("Ignoring LEGACY room link entry", entry);
```
turt2live: warn?

lib/SlackGhost.js
```diff
@@ -151,7 +152,7 @@ SlackGhost.prototype.uploadContent = function(file) {
     }).then((response) => {
         var content_uri = JSON.parse(response).content_uri;
 
-        console.log("Media uploaded to " + content_uri);
+        log.debug("Media uploaded to " + content_uri);
```
turt2live: info? Feels a bit less debug-y, although could really be argued either way

lib/SlackHookHandler.js
```diff
@@ -91,15 +92,15 @@ SlackHookHandler.prototype.startAndListen = function(port, tls_config) {
...
+    log.info("Received slack webhook " + method + " " + url + ": " + JSON.stringify(params));
 
     var main = this._main;
 
     var endTimer = main.startTimer("remote_request_seconds");
 
     var result = url.match(/^\/(.{32})(?:\/(.*))?$/);
     if (!result) {
-        console.log("Ignoring message with bad slackhook URL " + url);
+        log.info("Ignoring message with bad slackhook URL " + url);
```
turt2live: warn? warn is used for other 'Ignoring...' messages later on in the PR

lib/SlackHookHandler.js
```diff
@@ -121,7 +122,7 @@ SlackHookHandler.prototype.handle = function(method, url, params, response) {
     var room = main.getRoomByInboundId(inbound_id);
 
     if (!room) {
-        console.log("Ignoring message from unrecognised inbound ID: %s (%s.#%s)",
+        log.info("Ignoring message from unrecognised inbound ID: %s (%s.#%s)",
```
turt2live: warn? warn is used for other 'Ignoring...' messages later on in the PR

lib/SlackHookHandler.js
```diff
@@ -139,7 +140,7 @@ SlackHookHandler.prototype.handle = function(method, url, params, response) {
             (e) => {
                 endTimer({outcome: "fail"});
 
-                console.log("Failed: ", e);
+                log.error("Failed: ", e);
```
turt2live: again: what failed?

lib/SlackHookHandler.js
```diff
@@ -299,15 +300,15 @@ SlackHookHandler.prototype.replaceChannelIdsWithNames = function(message, token)
...
                 message.text = message.text.replace(/<#(\w+)\|?\w*?>/, "#" + response.channel.name);
             }
             else {
-                console.log("channels.info returned no result for " + id);
+                log.info("channels.info returned no result for " + id);
             }
             iteration++;
             }).catch((err) => {
-               console.log("Caught error " + err);
+               log.error("Caught error " + err);
```
turt2live: another helpful message

lib/SlackHookHandler.js
```diff
@@ -376,7 +376,7 @@ SlackHookHandler.prototype.replaceUserIdsWithNames = function(message, token) {
                 );
             }
         }).catch((err) => {
-            console.log("Caught error " + err);
+            log.error("Caught error " + err);
```
turt2live: helpful message

lib/Main.js
Half-Shot: It only happens if you try to PM the bridge, but it's rare enough for an info I guess.

lib/Main.js
Half-Shot: I'd call it an error since it shouldn't really ever happen. And it's only going to be emitted on startup.

lib/Main.js
Half-Shot: Same as above.

lib/SlackGhost.js
Half-Shot: I'm just going to leave this on debug on the basis that it's just telling us media got uploaded, which is what we would expect to happen and it doesn't really tell us a whole lot unless we are debugging an issue.

PR #95, opened 2018-10-05
add a dbdir config to customise location of the stores

Cadair:
Signed-off-by: Stuart Mumford <stuart@cadair.com>

### DIFF
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -50,10 +50,13 @@ function Main(config) {
     // So we can't create the StateLookup instance yet
     this._stateStorage = null;
 
+    var dbdir = config.dbdir || "";
     this._bridge = new Bridge({
         homeserverUrl: config.homeserver.url,
         domain: config.homeserver.server_name,
         registration: "slack-registration.yaml",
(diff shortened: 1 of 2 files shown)


lib/Main.js
```diff
@@ -50,10 +50,13 @@ function Main(config) {
     // So we can't create the StateLookup instance yet
     this._stateStorage = null;
 
+    var dbdir = config.dbdir || "";
```
Half-Shot: ``const`` please.

lib/Main.js
```diff
@@ -50,10 +50,13 @@ function Main(config) {
...
     this._bridge = new Bridge({
         homeserverUrl: config.homeserver.url,
         domain: config.homeserver.server_name,
         registration: "slack-registration.yaml",
+        userStore: dbdir + "user-store.db",
```
Half-Shot: I prefer https://nodejs.org/dist/latest-v10.x/docs/api/path.html#path_path_join_paths simply because it stops people forgetting a /

Half-Shot:
Thaaaaanks!

PR #96, opened 2018-10-05
Convert Riot Pills to Slack mentions

Cadair:
This is a replacement for #53

### DIFF
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -191,7 +191,7 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
-Main.prototype.getUsernamePrefix = function() {
+Main.prototype.getUserIDPrefix = function() {
     return this._config.username_prefix;
 };
 
--- a/lib/Main.js
+++ b/lib/Main.js
@@ -191,6 +191,10 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
+Main.prototype.getUsernamePrefix = function() {
+    return this._config.username_prefix;
+};
+
 Main.prototype.getTeamDomainForMessage = function(message) {
     if (message.team_domain) {
         return Promise.resolve(message.team_domain);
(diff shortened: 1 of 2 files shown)


lib/Main.js
```diff
@@ -191,6 +191,10 @@ Main.prototype.getBotIntent = function() {
     return this._bridge.getIntent();
 };
 
+Main.prototype.getUsernamePrefix = function() {
```
Half-Shot: This should probably be "userid" prefix.

lib/substitutions.js
```diff
@@ -134,6 +133,23 @@ var matrixToSlack = function(event, main) {
         if (event.content.msgtype == "m.emote") {
             string = "_" + string + "_";
         }
+
+        // replace riot "pill" behavior to "@" mention for slack users
+        var html_string = event.content.formatted_body;
```
Half-Shot: ``const`` please.

lib/substitutions.js
```diff
@@ -134,6 +133,23 @@ var matrixToSlack = function(event, main) {
...
+        if (undefined != html_string) {
+            var regex = new RegExp('<a href="https://matrix.to/#/@' +
```
Half-Shot: ``const`` please.

lib/substitutions.js
```diff
@@ -134,6 +133,23 @@ var matrixToSlack = function(event, main) {
...
+                                   main.getUsernamePrefix() +
+                                   '([^"]+)">([^<]+)</a>', "g");
+
+            var match = regex.exec(html_string);
```
Half-Shot: ``let`` please.

lib/substitutions.js
```diff
@@ -134,6 +133,23 @@ var matrixToSlack = function(event, main) {
...
+            while (match != null) {
+                // Extract the slack ID from the matrix id
+                var userid = match[1].split("_")[1].split(":")[0];
```
Half-Shot: ``const`` please.

Cadair:
What do you mean by config and schema update? This doesnt touch the config no?

Cadair:
The username_prefix option is already in the config and the [schema](https://github.com/matrix-org/matrix-appservice-slack/blob/master/config/slack-config-schema.yaml#L38). I assume it dosent need changing?

Half-Shot:
So it is, I thought you had added it. Looks good to me then :)

PR #98, opened 2018-10-05
Enable support for edits by copying it out of Gitter

Cadair:
Signed-off-by: Stuart Mumford <stuart@cadair.com>

closes #71 and closes #28

PR #99, opened 2018-10-07
Implement bridging of plain text matrix mentions

Cadair:
The Slack API removed the ability to just send slack `@Cadair` text and have slack convert it into a mention. This is an attempt to do this substitution automatically in the bridge.

~~I would like to remove the need for the `@` here as well, which I have an idea how to do, but I am a little concerned about the performance impacts of all this.~~

This now correctly bridges plain text matrix mentions as slack mentions. This means that all matrix clients should give the correct behaviour on slack, making the bridge truly transparent. :tada: 

This also builds on #98 because of many many conflicts.

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -192,9 +192,9 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
 
     return this._main.getOrCreateMatrixUser(message.user_id).then((user) => {
         substitutions.matrixToSlack(message, this._main).then(body => {
-            var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
+            const uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
 
-            var sendMessageParams = {
+            const sendMessageParams = {
                 method: "POST",
                 json: true,
                 uri: uri,
@@ -211,7 +211,7 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
 
             sendMessageParams.body.username = user.getDisplaynameForRoom(message.room_id);
 
-            var avatar_url = user.getAvatarUrlForRoom(message.room_id);
+            const avatar_url = user.getAvatarUrlForRoom(message.room_id);
 
             if (avatar_url && avatar_url.indexOf("mxc://") === 0) {
                 sendMessageParams.body.icon_url = this._main.getUrlForMxc(avatar_url);
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -192,8 +192,6 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
 
     return this._main.getOrCreateMatrixUser(message.user_id).then((user) => {
         substitutions.matrixToSlack(message, this._main).then(body => {
-            var body = substitutions.matrixToSlack(message, this._main);
-
             var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
 
             var sendMessageParams = {
@@ -219,8 +217,8 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
                 sendMessageParams.body.icon_url = this._main.getUrlForMxc(avatar_url);
             }
 
-        user.bumpATime();
-        this._matrixAtime = Date.now() / 1000;
+            user.bumpATime();
+            this._matrixAtime = Date.now() / 1000;
 
             return rp(sendMessageParams).then((res) => {
                 this._main.incCounter("sent_messages", {side: "remote"});
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -191,41 +191,43 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
     if (!this._slack_webhook_uri && !this._slack_bot_token) return Promise.resolve();
 
     return this._main.getOrCreateMatrixUser(message.user_id).then((user) => {
-        var body = substitutions.matrixToSlack(message, this._main);
+        substitutions.matrixToSlack(message, this._main).then(body => {
+            var body = substitutions.matrixToSlack(message, this._main);
 
-        var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
+            var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
 
-        var sendMessageParams = {
-            method: "POST",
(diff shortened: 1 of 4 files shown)


Cadair:
My concern with this as implemented is the potential effect on performance. This https://github.com/matrix-org/matrix-appservice-slack/blob/9975146be83d007949438205060c977ca5323108/lib/substitutions.js#L248-L254 
block of code is being executed for each message, irrespective of if there is a match or not (this was not the case when matching `@`), which means that this:
https://github.com/matrix-org/matrix-appservice-slack/blob/9975146be83d007949438205060c977ca5323108/lib/substitutions.js#L197-L210

function, which calls out to room state and the user store is also being hit on each message.

I don't know enough about the way the bridge lib handles caching of both the room state store or the user store to know if this is an issue.

Cadair:
This does also have the downside of being unable to handle the situation where you have two users on slack with the same display name. I guess the only way to semi-guard against this is to do some crazy regex on nicks of the form `Cadair (@slack_cadair_UDAS26GDC:testmatrix.home.cadair.com)` but I don't really fancy that frankly.

turt2live:
(merged develop to make review easier - don't mind me)

lib/BridgedRoom.js
```diff
@@ -191,41 +191,41 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
...
-            uri: uri,
-            body: body,
-        };
-        if (this._slack_bot_token) {
-            sendMessageParams.headers = {
-                Authorization: 'Bearer ' + this._slack_bot_token
+        substitutions.matrixToSlack(message, this._main).then(body => {
+            var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
+
+            var sendMessageParams = {
```
Half-Shot: :worried:

lib/BridgedRoom.js
```diff
@@ -191,41 +191,41 @@ BridgedRoom.prototype.onMatrixMessage = function(message) {
...
-            method: "POST",
-            json: true,
-            uri: uri,
-            body: body,
-        };
-        if (this._slack_bot_token) {
-            sendMessageParams.headers = {
-                Authorization: 'Bearer ' + this._slack_bot_token
+        substitutions.matrixToSlack(message, this._main).then(body => {
+            var uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
```
Half-Shot: :worried:

lib/substitutions.js
```diff
@@ -158,33 +158,139 @@ var matrixToSlack = function(event, main) {
...
-            image_url: url
-        }];
-    } else if (event.content.msgtype == "m.file" && event.content.url.indexOf("mxc://") === 0) {
-        var url = main.getUrlForMxc(event.content.url);
-        ret.text = "<" + url + "|" + string + ">";
-    }
+    // Note: This slower plainTextSlackMentions call is only used if there is not a pill in the message,
+    // meaning if we can we use the much simpler pill subs rather than this.
+    return plainTextSlackMentions(main, string, event.room_id).then(string => {
+        var ret = {
```
Half-Shot: :worried:

lib/substitutions.js
```diff
@@ -158,33 +158,139 @@ var matrixToSlack = function(event, main) {
...
+            username: event.user_id,
+            text: string,
+            link_names: 1  //This no longer works for nicks but is needed to make @channel work.
+        };
+        if (event.content.msgtype == "m.image" && event.content.url.indexOf("mxc://") === 0) {
+            var url = main.getUrlForMxc(event.content.url);
```
Half-Shot: :worried:

lib/substitutions.js
```diff
@@ -158,33 +158,139 @@ var matrixToSlack = function(event, main) {
...
+            delete ret.text;
+            ret.attachments = [{
+                fallback: string,
+                image_url: url
+            }];
+        } else if (event.content.msgtype == "m.file" && event.content.url.indexOf("mxc://") === 0) {
+            var url = main.getUrlForMxc(event.content.url);
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
+var subsitutions = require("../lib/substitutions");
+
+describe ("Plain Text Mentions", () => {
+    var displaymap = {Stuart: "U1111", Cadair: "U2222", "Stuart Mumford": "U3333"}
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    it("replace unique nick", () => {
+        var new_string = subsitutions.replacementFromDisplayMap("@Cadair", displaymap);
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+        expect(new_string).toBe("<@U2222>");
+    });
+    it("replace shortest non-unique", () => {
+        var new_string = subsitutions.replacementFromDisplayMap("@Stuart", displaymap);
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+        expect(new_string).toBe("<@U1111>");
+    });
+    it("replace longest non-unique", () => {
+        var new_string = subsitutions.replacementFromDisplayMap("@Stuart Mumford", displaymap);
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+        expect(new_string).toBe("<@U3333>");
+    });
+})
+
+describe ("Make First Word Map", () => {
+    var displaymap = {Stuart: "U1111", Cadair: "U2222", "Stuart Mumford": "U3333"}
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    var expectedfirstwords = {Stuart: [{Stuart: "U1111"}, {"Stuart Mumford": "U3333"}],
```
Half-Shot: :worried:

tests/test-subsitutions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+                              Cadair: [{Cadair: "U2222"}]};
+    it("Make the First Word Map", () => {
+        var firstwords = subsitutions.makeFirstWordMap(displaymap);
```
Half-Shot: :worried:

Half-Shot:
Thank you!

PR #105, opened 2018-10-18
Better to leave the text un-modified if we dont understand the emoji

Cadair:
Signed-off-by: Stuart Mumford <stuart@cadair.com>

### DIFF
--- a/lib/substitutions.js
+++ b/lib/substitutions.js
@@ -91,6 +91,9 @@ var slackToMatrix = function(body, file) {
         if (emojiChar !== undefined) {
             return emojiChar;
         }
+        else {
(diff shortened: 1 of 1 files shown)


lib/substitutions.js
```diff
@@ -91,6 +91,9 @@ var slackToMatrix = function(body, file) {
         if (emojiChar !== undefined) {
             return emojiChar;
         }
+        else {
+            return emojiKey;
+        }
         log.warn("Failed to recognise emoji sequence: " + emojiKey);
```
turt2live: actually wait, this logging won't be called anymore.

PR #109, opened 2018-10-22
Remove slack link escaping because slackdown does it better

Cadair:
I think this is the best solution, it fixes the pills bug for me, and also seems to not break normal links.

### DIFF
--- a/lib/SlackGhost.js
+++ b/lib/SlackGhost.js
@@ -162,8 +162,12 @@ SlackGhost.prototype.sendText = function(room_id, text) {
     // but the only parser we have for it is slackdown. However, Matrix expects
     // a variant of markdown that is in the realm of sanity. Currently text
     // will be slack's markdown until we've got a slack -> markdown parser.
+
+    //TODO: This is fixing plaintext mentions, but should be refactored. See issue #110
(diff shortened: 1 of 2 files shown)


Half-Shot:
Post me a screenshot of proof and I'll merge :p

Cadair:
Ok, so this is a fix to the pills bug, however, it exposes a new and interesting issue, in that the desired behaviour for user mentions for plain text and formatted parts of the message is different. If I were to `@Cadair` on slack we would want the plain text body to read `Cadair` and the formatted body to be `<a href="...">Cadair</a>` the current subs method (with this PR) of taking the slack message and doing the following transforms on it:

* `<@U1234>` ==> `<https://matrix.to/...|Cadair>` (In `BaseSlackHandler.js`)
* `<https://matrix.to/...|Cadair>` ==> `<a href="...">Cadair</a>` (Slackdown)

leads to the plain text body having  `<https://matrix.to/...|Cadair>` in it.

I am a little confused by the various locations we do the transformations throughout the pipe. I would think it would be worth refactoring the whole lot, so that the slack message goes into `subs.slackToMatrix` and comes out with a `body, formatted_body` pair. I.e does all the store, regex and slackdown subs?

PR #111, opened 2018-10-25
Fix the discrepancy between nicks and names in pills and mentions

Cadair:
This completely refactors `BaseSlackHandler.replaceUserIdsWithNames` to first check the UserStore for the display name and then fallback on querying slack for the display name. This means it should always be consistent with the nick of any user in the matrix room.

Signed off by: Stuart Mumford <stuart@cadair.com>

### DIFF
--- a/lib/BaseSlackHandler.js
+++ b/lib/BaseSlackHandler.js
@@ -98,14 +98,15 @@ BaseSlackHandler.prototype.replaceUserIdsWithNames = async function(message, tok
                 avatar_url: null});
             var room = main.getRoomBySlackChannelId(message.channel);
             display_name = await nullGhost.getDisplayName(id, room.getAccessToken());
-            user_id = main.getUserId(message.user_id, team_domain);
+            // If the user is not in the room, we cant pills them, we have to just plain text mention them.
+            message.text = message.text.replace(USER_ID_REGEX_FIRST, display_name);
         } else {
             display_name = users[0].display_name || user_id;
+            message.text = message.text.replace(
+                USER_ID_REGEX_FIRST,
+                `<https://matrix.to/#/${user_id}|${display_name}>`
+            );
         }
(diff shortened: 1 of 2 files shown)


lib/BaseSlackHandler.js
```diff
@@ -69,84 +70,43 @@ BaseSlackHandler.prototype.replaceChannelIdsWithNames = function(message, token)
...
-            if (result === undefined) {
-                return;
-            }
-            let name = user_id;
-            log.info(`${user_id} did ${result.length > 0 ? "not" : ""} an entry`);
-            if (result.length) {
-                // It's possible not to have a displayname set.
-                name = result[0].display_name || result[0].id;
-            }
+        var id = match[0].match(USER_ID_REGEX_FIRST)[1];
```
Half-Shot: VAR!

Cadair:
Alright @Half-Shot it's only been a month and a half, but I changed that `var` :rofl:

lib/BaseSlackHandler.js
```diff
@@ -69,84 +70,43 @@ BaseSlackHandler.prototype.replaceChannelIdsWithNames = function(message, token)
...
+        let user_id = main.getUserId(id, team_domain);
+
+        const store = main.getUserStore();
+        const users = await store.select({id: user_id});
+
+        if (users === undefined || !users.length) {
+            log.warn("Mentioned user not in store. Looking up display name from slack.");
+            // if the user is not in the store then we look up the displayname
+            const nullGhost = new slackGhost({main: main});
+            var room = main.getRoomBySlackChannelId(message.channel);
```
Half-Shot: VAR :cry:

Cadair: done :)

Half-Shot:
Oh, go on then!

PR #123, opened 2019-03-14
Remove hardcoding JPEG as filetype in thumbnail URIs

Berulacks:
It seems that Slack now _does_ pass on non-JPEGs as thumbnails. I wasn't able to send any file other than `jpeg`s, previously. This quick change removes the hard-coded ".jpeg" extensions from outgoing messages in favor of their reported file-type.

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -344,12 +344,11 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
                 let thumbnail_promise = Promise.resolve();
                 // Slack ain't a believer in consistency.
                 const thumb_uri = file.thumb_video || file.thumb_360;
-                if (thumb_uri) {
+                if (thumb_uri && file.filetype) {
                     thumbnail_promise = ghost.uploadContentFromURI(
                         {
-                            // Yes, we hardcode jpeg. Slack always use em.
-                            title: `${file.name}_thumb.jpeg`,
-                            mimetype: "image/jpeg",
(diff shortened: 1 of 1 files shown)


lib/BridgedRoom.js
```diff
@@ -344,12 +344,11 @@ BridgedRoom.prototype._handleSlackMessage = function(message, ghost) {
...
                 const thumb_uri = file.thumb_video || file.thumb_360;
-                if (thumb_uri) {
+                if (thumb_uri && file.filetype) {
                     thumbnail_promise = ghost.uploadContentFromURI(
                         {
-                            // Yes, we hardcode jpeg. Slack always use em.
-                            title: `${file.name}_thumb.jpeg`,
-                            mimetype: "image/jpeg",
+                            title: `${file.name}_thumb.${file.filetype}`,
+                            mimetype: "image/" + file.filetype,
```
vurpo:
```suggestion
                            mimetype: file.mimetype,
```

`file.mimetype` would probably lead to more predictable and correct results here. (This is the way I implemented this change on our instance of the bridge at hacklab.fi)

olmari: @vurpo method we also tested and confirmed working with multiple diffirent image formats on said hacklab.fi bridge.

Berulacks: Makes sense. Cheers!

Berulacks:
Signed-off-by: Derin Yarsuvat <derin@ml1.net>

PR #120, opened 2019-01-12
Add basic support (via fallbacks) for attachments.

Cadair:
Signed-off-by: Stuart Mumford <stuart@cadair.com> 

This is a mitigation for #36 and also includes improvements to how bot users are represented in matrix.

PR #126, opened 2019-05-03
Slack attachments with requested changes

umitalp:
PR #120  With Requested changes @Half-Shot @jbrinksmeier

### DIFF
--- a/lib/SlackEventHandler.js
+++ b/lib/SlackEventHandler.js
@@ -159,12 +159,6 @@ SlackEventHandler.prototype.handleMessageEvent = async function (params) {
         return;
     }
 
-    // If it's a message_deleted event we should redact it on the matrix side.
-    if (params.event.subtype === 'message_deleted') {
-        // TODO: Implement this, but it needs event id lookup.
-        return;
-    }
-
     // Only count received messages that aren't self-reflections
     this._main.incCounter("received_messages", {side: "remote"});
 
@@ -182,7 +176,7 @@ SlackEventHandler.prototype.handleMessageEvent = async function (params) {
         for (var attachment of params.event.attachments) {
             msg.text = attachment.fallback;
             msg = await this.doChannelUserReplacements(msg, token);
-            await room.onSlackMessage(msg);
+            return await room.onSlackMessage(msg);
         }
         if (params.event.text == '') {
             return;
@@ -196,7 +190,8 @@ SlackEventHandler.prototype.handleMessageEvent = async function (params) {
         // TODO(paul): When I started looking at this code there was no lookupAndSendMessage()
         //   I wonder if this code path never gets called...?
         // lookupAndSendMessage(params.channel_id, params.timestamp, intent, roomID, token);
-        log.warn("Did not understand event: ", JSON.stringify(message));
+        log.warn("Did not understand event: ", message);
+        log.debug("Did not understand event: ", message);
         return;
     }
 
@@ -214,7 +209,7 @@ SlackEventHandler.prototype.handleMessageEvent = async function (params) {
         // (because we don't have a master token), but it has text,
         // just send the message as text.
         log.warn("no slack token for " + room.getSlackTeamDomain() || room.getSlackChannelId());
-        await room.onSlackMessage(msg);
+        return await room.onSlackMessage(msg);
     }
 
     if (msg.subtype === "file_share" && msg.file) {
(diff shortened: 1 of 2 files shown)


lib/SlackEventHandler.js
```diff
@@ -140,13 +145,24 @@ SlackEventHandler.prototype.handleChannelRenameEvent = function (params) {
...
+        return;
+    }
+
+    // If it's a ephemeral event we don't have a way of relaying it.
+    if (params.event.is_ephemeral) {
+        return;
+    }
+
+    // If it's a message_deleted event we should redact it on the matrix side.
+    if (params.event.subtype === 'message_deleted') {
```
Half-Shot: Do we need this, @Cadair has a seperate PR for this so it feels like it's a pointless if loop

lib/SlackEventHandler.js
```diff
@@ -161,7 +177,28 @@ SlackEventHandler.prototype.handleMessageEvent = function (params) {
...
+        msg.text = params.event.text;
+    }
+
+    const processMsg = msg.text || msg.files != undefined || msg.subtype === 'message_changed';
+
+    if (!processMsg) {
+        // TODO(paul): When I started looking at this code there was no lookupAndSendMessage()
+        //   I wonder if this code path never gets called...?
+        // lookupAndSendMessage(params.channel_id, params.timestamp, intent, roomID, token);
+        log.warn("Did not understand event: ", JSON.stringify(message));
```
Half-Shot: Shouldn't need to stringify, logger does that automatically.

Also make the `message` get logged as debug, but the error as warn because otherwise the logs will be incredibly bloated.

lib/SlackEventHandler.js
```diff
@@ -177,43 +214,26 @@ SlackEventHandler.prototype.handleMessageEvent = function (params) {
         // (because we don't have a master token), but it has text,
         // just send the message as text.
         log.warn("no slack token for " + room.getSlackTeamDomain() || room.getSlackChannelId());
-
-        return (processMsg) ? room.onSlackMessage(msg) : Promise.resolve();
+        await room.onSlackMessage(msg);
```
Half-Shot: Is it intentional to no longer return the result of this promise?

lib/SlackEventHandler.js
```diff
@@ -177,43 +214,26 @@ SlackEventHandler.prototype.handleMessageEvent = function (params) {
...
-    } else {
-        result = Promise.resolve();
     }
 
-    return result.then(() => msg)
-        .then((msg) => this.replaceChannelIdsWithNames(msg, token))
-        .then((msg) => this.replaceUserIdsWithNames(msg, token))
-        .then((msg) => room.onSlackMessage(msg));
+    msg = await this.doChannelUserReplacements(msg, token);
+    await room.onSlackMessage(msg);
```
Half-Shot: Same as above.

lib/SlackGhost.js
```diff
@@ -60,34 +60,44 @@ SlackGhost.prototype.update = function(message, room) {
...
+
+    }
+    return;
 };
 
 SlackGhost.prototype.lookupAvatarUrl = function(user_id, token) {
     return this.lookupUserInfo(user_id, token).then((user) => {
-        if (!user || !user.profile) return;
+        if (!user || !user.profile) {
+            log.debug("Can't read user profile, so cannot lookup avatar url");
```
Half-Shot: Log the user we failed to get the profile for?

lib/SlackGhost.js
```diff
@@ -60,34 +60,44 @@ SlackGhost.prototype.update = function(message, room) {
...
             this._display_name = display_name;
             return this._main.putUserToStore(this);
         });
-    })
+
+    }
+    return;
 };
 
 SlackGhost.prototype.lookupAvatarUrl = function(user_id, token) {
```
Half-Shot: The other functions are async, but this isn't? It looks like a good candidate.

lib/SlackEventHandler.js
```diff
@@ -177,43 +214,26 @@ SlackEventHandler.prototype.handleMessageEvent = function (params) {
...
-                    if (file) {
-                        msg.file = file;
-                    }
-
-                    return this.fetchFileContent(msg.file, token)
-                        .then((content) => {
-                            msg.file._content = content;
-                        });
-                })
+            const file = this.enablePublicSharing(msg.file, room.getSlackUserToken());
```
Half-Shot: This is a promise, so await it?

lib/SlackEventHandler.js
umitalp: Should `message` to be `msg` or `params` instead?

Half-Shot:
Thanks for your work @umitalp, this has been merged in #147 :)

PR #130, opened 2019-05-22
Add support for Edits

Cadair:
This adds support for the edits in riot develop bidirectionally with slack.

Slack -> Matrix edits still display the rich fallback to clients which don't speak the new edits.

builds on #129 

Signed Off By Stuart Mumford (stuart@cadair.com)

PR #129, opened 2019-05-21
Implement message deletion

Cadair:
fixes #70 and fixes #100 (although not all the other things that this enables)

PR #131, opened 2019-05-22
Reactions bridging

Cadair:
~builds on #130 which builds on #129~

TODO:

- [ ] Need to use aggregations API to handle slack -> matrix reaction removal.


Cadair:
I am not sure I can quite get the aggregations stuff in synapse to play ball for this, but I am trying.

Cadair:
it seems matrix -> slack reaction removal might be broken.

Cadair:
So it seems that slack has alternative names for a bunch of emoji i.e `:+1:` and `:thumbsup:` and this works with `:+1:` but not `:thumbsup:`. 


AHGHGGGGHHHHHHHLLLLLL

Cadair:
Fixed.

Cadair:
this is now based on #153

lib/BridgedRoom.js
```diff
@@ -190,6 +191,58 @@ BridgedRoom.prototype.toEntry = function() {
     return entry;
 };
 
+BridgedRoom.prototype.onMatrixReaction = async function(message) {
+    if (!this._slack_bot_token) return Promise.resolve();
```
Half-Shot: This is an async function, just return

lib/BridgedRoom.js
```diff
@@ -389,6 +446,68 @@ BridgedRoom.prototype.onSlackMessage = async function(message) {
...
+};
+
+BridgedRoom.prototype.getMatrixReactionsBySenderByKey = async function(eventId, sender, key) {
+    // The objective of this function is to hit the new
+    // rooms/roomId/aggregations endpoint for all the m.reaction events with the
+    // specified key.
+    // Then we walk that list and find the one with the sender we care about, and return that event.
+
+    // TODO: Not sure this works, having trouble testing it against a sufficiently new synapse.
+    return;
```
Half-Shot: Please just remove the func if it doesn't work. I'm not sure it's worth bloating the code base out with future things if the spec isn't even merged yet.

lib/SlackEventHandler.js
```diff
@@ -146,7 +152,7 @@ SlackEventHandler.prototype.doChannelUserReplacements = async function(msg, text
  * @param {string} params.event.ts The unique (per-channel) timestamp
  */
 SlackEventHandler.prototype.handleMessageEvent = async function (params) {
-    var room = this._main.getRoomBySlackChannelId(params.event.channel);
+    var room = this._main.getRoomBySlackChannelId(params.event.channel || params.event.item.channel);
```
Half-Shot: VAR

Half-Shot:
FYI, we decided to not do redactions of reactions slack -> matrix because it involves having to listen for your own reactions over the events API and it can wait for a future PR.

PR #132, opened 2019-05-25
Threads replies bridging

### LINKED ISSUE #45: How to deal with threads
Today I kept seeing messages in Matrix that I wasn't seeing in the actual Slack web client.  I pulled out my hair for a while trying to figure out what was going on, thinking perhaps someone had blocked me personally or something, before eventually finding the messages in a thread from days earlier that I wasn't following.  I understand if Matrix doesn't support threaded conversations (yet, I hope!), but it would be extremely helpful if there was some kind of indicator that a message was a part of a threaded conversation, even something as simple as a ↳ before the message or something might be good enough.

### LINKED ISSUE #57: Slack option “Also send to #channel-name.” prevents message transmission
Hi,

if Slack option [“Also send to #channel-name.”](https://get.slack.help/hc/en-us/articles/115000769927-Message-threads#send-a-reply-to-the-entire-channel) is activated message delivery to bridged Matrix room does not work. 
So a message posted in a Slack thread with activated “Also send to #channel-name.” doesn't appear in Matrix without any error on both sides.

Thx for Matrix and the Slack bridge, hope this helps to improve the bridging mode!
anonybot

Cadair:
Initial work on threads to replies.

So far only matrix replies get sent to slack threads, and I need to add stripping the reply fallbacks.

builds on ~#129~ and ~#130~

Sorta fixes #45 and far as I can tell fixes #57 

Signed off By: Stuart Mumford (stuart@cadair.com)

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -323,9 +323,9 @@ BridgedRoom.prototype.onMatrixMessage = async function(message) {
     const store = this._main.getEventStore();
 
     const user = await this._main.getOrCreateMatrixUser(message.user_id);
-    const stripped_message = this.stripMatrixReplyFallback(message);
-    log.error("aa", stripped_message);
-    const body = await substitutions.matrixToSlack(stripped_message, this._main);
+
+    message = this.stripMatrixReplyFallback(message);
+    const body = await substitutions.matrixToSlack(message, this._main);
     const uri = (this._slack_bot_token) ? "https://slack.com/api/chat.postMessage" : this._slack_webhook_uri;
 
     const sendMessageParams = {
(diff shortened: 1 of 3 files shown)


Cadair:
@Half-Shot rebased on #130.

lib/BridgedRoom.js
```diff
@@ -264,10 +264,67 @@ BridgedRoom.prototype.onMatrixEdit = async function(message) {
...
+    // Extract the referenced event
+    if (!message["content"]) return message.event_id;
+    if (!message["content"]["m.relates_to"]) return message.event_id;
+    if (!message["content"]["m.relates_to"]["m.in_reply_to"]) return message.event_id;
+    const parentEventId = message["content"]["m.relates_to"]["m.in_reply_to"]["event_id"];
+    if (!parentEventId) return message.event_id;
+
+    // Get the previous event
+    const intent = this._main.getBotIntent();
+    let nextEvent = await intent.getClient().fetchRoomEvent(message.room_id, parentEventId);
```
Half-Shot: const pls

lib/SlackGhost.js
```diff
@@ -120,7 +120,7 @@ SlackGhost.prototype.lookupUserInfo = function(user_id, token) {
         json: true,
     }).then((response) => {
         if (!response.user || !response.user.profile) {
-            log.error("Failed to get user profile", response);
+            log.error("Failed to get user profile for ${user_id}", response);
```
Half-Shot: This should be in backticks

Half-Shot:
LGTM

PR #143, opened 2019-06-25
Remove duplication of registration path and tidy example config

Cadair:
We really need to update our docs in general but hopefully mentioning these two options in the example config is at least a step in the right direction.

@kingoftheconnors @spantaleev

Signed off By: Stuart Mumford (stuart@cadair.com)

### DIFF
--- a/app.js
+++ b/app.js
@@ -21,10 +21,10 @@ const cli = new Cli({
         reg.addRegexPattern("users", "@" + config.username_prefix + ".*", true);
         callback(reg);
     },
-    run: function(port, config) {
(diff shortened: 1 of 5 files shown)


Cadair:
Ok, had a change around here, now the registration file is only handled by the CLI. So it's not in the config and the correct way to set it is with the `--file` when starting `app.js`.

Cadair:
It shouldn't. The default for the cli arg is `"slack-registration.yaml"` which is what the bridge defaulted to as well, so if it's not specified it should just all null out to the old values.

PR #149, opened 2019-07-03
Fix mentions

### LINKED ISSUE #148: Slack -> matrix pills are broken on develop

Cadair:
Signed Off By: Stuart Mumford (stuart@cadair.com)

fixes #148

### DIFF
--- a/lib/substitutions.js
+++ b/lib/substitutions.js
@@ -171,26 +171,25 @@ const matrixToSlack = async function(event, main) {
 
     // Note: This slower plainTextSlackMentions call is only used if there is not a pill in the message,
     // meaning if we can we use the much simpler pill subs rather than this.
-    return plainTextSlackMentions(main, string, event.room_id).then(string => {
-        const ret = {
-            username: event.user_id,
-            text: string,
-            link_names: 1  //This no longer works for nicks but is needed to make @channel work.
-        };
-        if (event.content.msgtype == "m.image" && event.content.url.indexOf("mxc://") === 0) {
-            const url = main.getUrlForMxc(event.content.url);
-            delete ret.text;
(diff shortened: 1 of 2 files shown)


lib/substitutions.js
```diff
@@ -69,9 +69,6 @@ var slackToMatrix = function(body, file) {
         }
     }
 
-    // remove <> from links, e.g:
-    body = body.replace(/<((https?:\/\/)?[^>]+?)>/g, '$1');
```
Half-Shot: oh?

lib/substitutions.js
```diff
@@ -188,19 +184,18 @@ var htmlEscape = function(s) {
...
-        .then(users => {
-            const displaymap = {};
-            users.forEach(user => {
-                if (user && user[0]) {
-                    displaymap[user[0].display_name] = user[0].id.split("_")[2].split(":")[0];
-                }
-            });
-            return displaymap;
-        });
+    //TODO: why is this not getting a list of users anymore?
```
Half-Shot: I don't think this is true anymore?

lib/substitutions.js
```diff
@@ -110,7 +107,7 @@ var getSlackFileUrl = function(file) {
  * @param {Main} main the toplevel main instance
  * @return An object which can be posted as JSON to the Slack API.
  */
-var matrixToSlack = function(event, main) {
+var matrixToSlack = async function(event, main) {
```
Half-Shot: ### var

lib/substitutions.js
Cadair: Yeah I added this because I thought its absence was causing problems, but it seems that I introduced this bug instead.

PR #153, opened 2019-07-05
Bug fixes for develop!

### LINKED ISSUE #150: Username replacement in Edits and threads is broken
i.e on slack edit a message and add a mention and it dosen't render correctly on matrix.

### LINKED ISSUE #151: Cross team matrix -> slack room substitutions dosen't work
We should only transform bridged rooms to slack channels if they are part of the same team as the channel we are sending the message to. Otherwise they render as `#unknown-channel` on slack and are therefore useless.

Cadair:
fixes #150 and also fixes #151 

Sorry this ended up being way more of an undertaking than I hoped.

Signed off by: Stuart Mumford (stuart@cadair.com)

### DIFF
--- a/lib/BaseSlackHandler.js
+++ b/lib/BaseSlackHandler.js
@@ -52,7 +52,7 @@ BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, 
     var main = this._main;
 
     if (text === undefined) {
-        return message;
+        return text;
     }
 
     // match all channelIds
@@ -95,7 +95,7 @@ BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, 
 
 BaseSlackHandler.prototype.replaceUserIdsWithNames = async function(message, text, token) {
     if (text === undefined) {
-        return message;
+        return text;
     }
 
     var main = this._main;
--- a/lib/BaseSlackHandler.js
+++ b/lib/BaseSlackHandler.js
@@ -48,15 +48,15 @@ BaseSlackHandler.prototype.getSlackRoomNameFromID = async function(id, token) {
     return id;
 };
 
-BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, token) {
+BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, text, token) {
     var main = this._main;
 
-    if (message.text === undefined) {
+    if (text === undefined) {
         return message;
     }
 
     // match all channelIds
-    var testForName = message.text.match(CHANNEL_ID_REGEX);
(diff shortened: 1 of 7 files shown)


Cadair:
Have now tested the second commit, it appears to work.

lib/BaseSlackHandler.js
```diff
@@ -48,15 +48,15 @@ BaseSlackHandler.prototype.getSlackRoomNameFromID = async function(id, token) {
     return id;
 };
 
-BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, token) {
+BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, text, token) {
```
Half-Shot: is message still used? Also you are returning message but editing text?

lib/BaseSlackHandler.js
```diff
@@ -48,15 +48,15 @@ BaseSlackHandler.prototype.getSlackRoomNameFromID = async function(id, token) {
...
     var main = this._main;
 
-    if (message.text === undefined) {
+    if (text === undefined) {
         return message;
     }
 
     // match all channelIds
-    var testForName = message.text.match(CHANNEL_ID_REGEX);
+    var testForName = text.match(CHANNEL_ID_REGEX);
```
Half-Shot: [var](https://media.tenor.com/images/50c5ec11ca69d065be8715b9950f41eb/tenor.gif)

lib/BaseSlackHandler.js
Cadair: Yeah, message is being used to lookup teams and senders and stuff. It should be returning text.

lib/BaseSlackHandler.js
Cadair: Do you just want a bigger diff to update to ts? :p

lib/BaseSlackHandler.js
```diff
@@ -48,15 +48,15 @@ BaseSlackHandler.prototype.getSlackRoomNameFromID = async function(id, token) {
     return id;
 };
 
-BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, token) {
+BaseSlackHandler.prototype.replaceChannelIdsWithNames = async function(message, text, token) {
     var main = this._main;
 
-    if (message.text === undefined) {
+    if (text === undefined) {
         return message;
```
Half-Shot: should this be `text`?

lib/BaseSlackHandler.js
Half-Shot: *grumbles* okay fine, I already fixed this in TS

PR #156, opened 2019-07-11
teams.db path should use dbdir

vrutkovs:
The bot won't accept the room invite if current working dir is not writable (e.g. in k8s environment). In this case `dbdir` has to be set.

This PR ensures the bridge won't crash when `dbdir` is set, as previously only local `teams.db` path was used

Signed-off-by: Vadim Rutkovsky <vrutkovs@redhat.com>


Half-Shot:
This looks great, thanks! Can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off? (I haven't copied the sign off rules to this repo yet, my bad).

PR #164, opened 2019-07-19
RTM API support

Half-Shot:
~~(This is currently using `hs/typescript` as a base for easy reading, but should be merged against `develop` eventually)~~

(This was meant to be a draft, but I failed to click the button)

This PR adds support for using the realtime messaging API for incoming slack messages. This means that rather than using HTTP push for events, they will come across webhooks which should be quicker, easier to setup and allows for special events like typing notifications.

This feature is also a hard blocker for things like deployment to Modular, because it makes setting up per-team events push vastly easier (events API can only push to one host and will push events from all teams, whereas this API can pull from any host and limit it to one team). It also means we don't need to pop open any ports.


src/SlackHandler.ts
```diff
@@ -0,0 +1,47 @@
+/**
```
Cadair: can you rename this file to `SlackRTMHandler`?

Cadair:
I wonder if Mattermost has a RTM api which is (mostly) slack equivalent? If it does we could consider dropping webhooks and events completely?

PR #167, opened 2019-07-19
Verify that the ID exists before storing it in the displaymap

### LINKED ISSUE #166: "the" in a Matrix room comes out as "@undefined" on the Slack side
Sometimes the word "the" on the Matrix side is replaced by "@undefined" in the Slack channel.

There is a Slack user whose display name starts with "the". In very limited testing, the issue did occur in a channel that user is in, but was not able to reproduce in a channel which that user is not in.

If it matters, this is Synapse 0.99.5.2 running on FreeBSD 12.  Please let me know if you need any additional information.

Half-Shot:
This fixes #166

### DIFF
--- a/lib/substitutions.js
+++ b/lib/substitutions.js
@@ -181,7 +181,7 @@ async function getDisplayMap(main, room_id) {
     users = users.filter((u) => u && u[0]);
     users.forEach(user => {
         // The format is @prefix%nospace%domain_userid:homeserver_domain
-        const localpart = user[0].id.split(":")[0].substr(this.main.getUserIdPrefix() + 1);
+        const localpart = user[0].id.split(":")[0].substr(this.main.getUserIdPrefix().length + 1);
         const slackId = localpart.split("_")[1];
         if (!slackId) {
             return;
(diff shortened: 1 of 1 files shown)


lib/substitutions.js
```diff
@@ -178,10 +178,15 @@ async function getDisplayMap(main, room_id) {
     let users = await main.listGhostUsers(room_id);
     users = await Promise.all(users.map(user => store.select({id: user})));
     const displaymap = {};
+    users = users.filter((u) => u && u[0]);
     users.forEach(user => {
-        if (user && user[0]) {
-            displaymap[user[0].display_name] = user[0].id.split("_")[2].split(":")[0];
+        // The format is @slack_domain_userid:homeserver_domain
+        const localpart = user[0].id.split(":")[0].replace("@slack", "").replace("@_slack", "");
```
Cadair: is the prefix here not customisable? I thought this was a variable on main somewhere?

Half-Shot: ...Yes you are right. For whatever reason this function isn't aware of that

Cadair: It could be though? I thought these had a reference to main?

Cadair: You should use [`main.getUserIDPrefix`](https://github.com/matrix-org/matrix-appservice-slack/blob/master/lib/Main.js#L206) here right?

Half-Shot:
Going to merge it then.

PR #169, opened 2019-07-22
Correctly add reply fallbacks on messages in threads, and edits in threads

Cadair:
This turned out to be much harder because of edits, but it works now.

### DIFF
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -290,8 +290,9 @@ BridgedRoom.prototype.onMatrixEdit = async function(message) {
     const event = await eventStore.getEntryByMatrixId(message.room_id, message.content['m.relates_to'].event_id);
 
     // re-write the message so the matrixToSlack converter works as expected.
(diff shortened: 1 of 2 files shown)


lib/BridgedRoom.js
```diff
@@ -540,11 +547,33 @@ BridgedRoom.prototype._handleSlackMessage = async function(message, ghost) {
...
         const prev_event = await eventStore.getEntryByRemoteId(slackRoomID, message.previous_message.ts);
+
+        // If this edit is in a thread we need to inject the reply fallback, or
+        // non-reply supporting clients will no longer show it as a reply.
+        let body = ghost.prepareBody(outtext);
+
+        let new_body = ghost.prepareBody(new_message_rich);
+        let new_formatted_body = ghost.prepareFormattedBody(new_message_rich);
+        if (message.message.thread_ts !== undefined) {
+            log.error("booo");
```
Half-Shot: :ghost: can we make this more descriptive or remove it.

PR #174, opened 2019-07-24
Correctly bookeep threads when we reply on matrix.

Cadair:
replaces #172

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /usr/src/app
 
 COPY . /usr/src/app
 
-RUN npm install --only=production
+RUN apk add git && npm install --only=production
 
 EXPOSE 9898
 EXPOSE 5858
--- a/lib/BridgedRoom.js
+++ b/lib/BridgedRoom.js
@@ -412,10 +412,11 @@ BridgedRoom.prototype.onMatrixMessage = async function(message) {
     sendMessageParams.body.username = user.getDisplaynameForRoom(message.room_id);
 
     const reply = await this.findParentReply(message);
+    let parentEvent = undefined;
     if (reply !== message.event_id) {
         // We have a reply
-        const parentStoredEvent = await store.getEntryByMatrixId(message.room_id, reply);
-        sendMessageParams.body.thread_ts = parentStoredEvent.remoteEventId;
+        parentEvent = await store.getEntryByMatrixId(message.room_id, reply);
+        sendMessageParams.body.thread_ts = parentEvent.remoteEventId;
     }
 
     const avatar_url = user.getAvatarUrlForRoom(message.room_id);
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /usr/src/app
 
 COPY . /usr/src/app
 
-RUN npm install --only=production
+RUN apk add git && npm install --only=production
```
Half-Shot: huh?

lib/BridgedRoom.js
```diff
@@ -412,10 +412,11 @@ BridgedRoom.prototype.onMatrixMessage = async function(message) {
     sendMessageParams.body.username = user.getDisplaynameForRoom(message.room_id);
 
     const reply = await this.findParentReply(message);
+    let parentEvent = undefined;
```
Half-Shot: It's implicit

```suggestion
    let parentEvent;
```

lib/BridgedRoom.js
```diff
@@ -412,10 +412,11 @@ BridgedRoom.prototype.onMatrixMessage = async function(message) {
...
     if (reply !== message.event_id) {
         // We have a reply
-        const parentStoredEvent = await store.getEntryByMatrixId(message.room_id, reply);
-        sendMessageParams.body.thread_ts = parentStoredEvent.remoteEventId;
+        parentEvent = await store.getEntryByMatrixId(message.room_id, reply);
+        sendMessageParams.body.thread_ts = parentEvent.remoteEventId;
```
Half-Shot: What happens if `parentEvent` is undefined?

lib/BridgedRoom.js
```diff
@@ -436,6 +437,16 @@ BridgedRoom.prototype.onMatrixMessage = async function(message) {
         // Add this event to the event store
         const event = new StoredEvent(message.room_id, message.event_id, this._slack_channel_id, res.ts);
         store.upsertEvent(event);
+
+        // If this message is in a slack thread we need to append this message to the end of the thread list.
+        if (parentEvent !== undefined) {
```
Half-Shot:
```suggestion
        if (parentEvent) {
```

Dockerfile
Cadair: It needed git to install your slackdown fork :man_shrugging:

lib/BridgedRoom.js
Cadair: Then something has gone horribly wrong and we don't have the event that's being replied to?

Dockerfile
Half-Shot: Stealing this for develop

lib/BridgedRoom.js
Half-Shot: We should definitely give up at that point and throw an error then, we can't assume we have it.

PR #185, opened 2019-08-28
Use Slack's library for calling HTTP endpoints

Half-Shot:
Sorry about the size of this one, but it was sort of unavoidable.

This PR changes the bridge significantly so that we now rely on `@slack/web-api` rather than doing HTTP calls ourselves, which reduces the amount of maintenance needed somewhat. API responses now also are typed and based off the slack documentation.

This PR can be treated separate from #164, but should be merged in beforehand ideally.

image

- pictured: reviewer wondering why these PRs are so massive

### DIFF
--- a/src/BaseSlackHandler.ts
+++ b/src/BaseSlackHandler.ts
@@ -19,6 +19,8 @@ import * as rp from "request-promise-native";
 import { Main } from "./Main";
 import { SlackGhost } from "./SlackGhost";
 import { default as subs } from "./substitutions";
+import { WebClient } from "@slack/web-api";
+import { FilesSharedPublicURLResponse, ConversationsInfoResponse } from "./SlackResponses";
 
 const log = Logging.get("BaseSlackHandler");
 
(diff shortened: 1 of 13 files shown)


Half-Shot:
(oh it broke, one second)

src/BaseSlackHandler.ts
```diff
@@ -104,41 +106,47 @@ export interface ISlackFile {
...
             return;
         }
-        text = await this.replaceChannelIdsWithNames(msg, text, token);
-        return await this.replaceUserIdsWithNames(msg, text, token);
+        text = await this.replaceChannelIdsWithNames(msg, text, webClient);
+        return await this.replaceUserIdsWithNames(msg, text);
     }
 
-    public async replaceChannelIdsWithNames(message: ISlackMessage, text: string, token: string) {
+    public async replaceChannelIdsWithNames(message: ISlackMessage, text: string, slackClient: WebClient) {
```
Cadair: why the mixing of `webClient` and `slackClient` shouldn't we pick one?

anoadragon453:
image

oh god why

> +455 −1,521

src/BaseSlackHandler.ts
Half-Shot: oh yeah, I think I was being indecisive. probably should use slackClient

Half-Shot:
>With all these changes the bridge is going to need significant testing before release.

This is 100% the plan.

PR #188, opened 2019-09-01
Slack -> Matrix topic changes

Half-Shot:
This is a simple change to set topics of rooms when slack sends us a topic update. Up for discussion is *when* should we apply topic changes. E.g. some matrix rooms bridged to slack would like the matrix topic to be authoritative.


Cadair:
Slack also has some kind of Room description field (which you can set at creation) which isn't the same as the topic.

I think this really needs to be opt-in on a room by room basis (and support sending updates in the opposite direction).

turt2live:
@Half-Shot CI is failing here

Half-Shot:
>Slack also has some kind of Room description field (which you can set at creation) which isn't the same as the topic.

This is the 'purpose' right? I am wondering if we should put that before or after the topic, or not at all.

PR #192, opened 2019-09-02
Fix passing registration in

Cadair:


src/app.ts
```diff
@@ -33,10 +33,10 @@ const cli = new Cli({
         reg.addRegexPattern("users", `@${config.username_prefix}.*`, true);
         callback(reg);
     },
-    run(port: number, config: IConfig) {
+    run(port: number, config: IConfig, registration) {
```
Half-Shot: `any` here please

src/Main.ts
```diff
@@ -105,7 +105,7 @@ export class Main {
     // track which teams are using the rtm client.
     private rtmTeams: Set<string> = new Set();
 
-    constructor(public readonly config: IConfig) {
+    constructor(public readonly config: IConfig, registration) {
```
Half-Shot: `any` here please

PR #194, opened 2019-09-02
Correctly bookeep threads when we reply on matrix

Cadair:
This is the ts port of #174


src/BridgedRoom.ts
```diff
@@ -379,9 +379,10 @@ export class BridgedRoom {
         };
 
         const reply = await this.findParentReply(message);
+        let parentStoredEvent;
```
Cadair: I figured out that this has the type `EventEntry` but I don't know how to reference that type (linter was saying it was undefined)

Half-Shot:
```suggestion
        let parentStoredEvent: EventEntry;
```

also import `import { EventEntry } from "./datastore/Models";`

Cadair: Having done this my linter tells me, `Type EventEntry | null is not assignable to type EventEntry` on the line below. What do I do to allow it to do both?

Half-Shot:
There appears to be other changes in here too, but I'll allow it.

PR #195, opened 2019-09-02
Fix race when trying to create more than one RTM client for a team

Half-Shot:
This can happen when multiple channels belong to the same team and try to call the start function. This code returns early and stores a promise to ensure that future calls will await the promise.

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -400,6 +400,8 @@ export class Main {
     public async addBridgedRoom(room: BridgedRoom) {
         this.rooms.push(room);
 
+        this.roomsByMatrixRoomId[room.MatrixRoomId] = room;
+
         if (room.SlackChannelId) {
             this.roomsBySlackChannelId[room.SlackChannelId] = room;
         }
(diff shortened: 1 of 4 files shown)


src/Main.ts
```diff
@@ -750,11 +753,16 @@ export class Main {
 
         await Promise.all(entries.map(async (entry) => {
             const hasToken = entry.remote.slack_team_id && entry.remote.slack_bot_token;
-            const cli = !hasToken ? undefined : await this.createOrGetTeamClient(
-                entry.remote.slack_team_id!, entry.remote.slack_bot_token!);
+
+            let cli: WebClient|undefined;
+            try {
+                cli = !hasToken ? undefined : await this.createOrGetTeamClient(
```
benparsons: Is this convention ok? Better to lose the `!` and swap the statements?

src/SlackHookHandler.ts
```diff
@@ -104,7 +104,7 @@ export class SlackHookHandler extends BaseSlackHandler {
                         this.eventHandler.onVerifyUrl(eventPayload.challenge!, eventsResponse);
                         // if RTM is enabled, don't handle this event. We should have assigned a RTM client to the team.
                     } else if (eventPayload.type === "event_callback" &&
-                               this.main.teamIsUsingRtm(eventPayload.team_id)
+                               !this.main.teamIsUsingRtm(eventPayload.team_id)
```
benparsons: This took a few extra seconds to read - is the use of RTM clients for teams documented/commented anywhere? Not clear why we only handle events in the absence of RTM.

Half-Shot: Ah, will document.

src/Main.ts
benparsons: Improved, thanks!

src/SlackHookHandler.ts
Half-Shot: I've documented, but to explain it here as well. RTM is an alternative to the events API that pushes the same data, but down websockets. We've left the events API code in for legacy reasons, but that means we need to ensure we do not handle both events coming in from the HookHandler and the RTMHandler for the same team.

benparsons: Improved comments are good + clear, thanks

PR #200, opened 2019-09-04
Puppeting support

Half-Shot:
This PR makes a lot of changes in order to support puppeting:

- Creates a new class `SlackClientFactory` to unify the creation of clients for tokens
- Extends the datastore so that it can store puppet tokens.
- Extends the oauth system to facilitate slack user oauth (instead of just bot oauth).
- Support sending messages from a user's account
- Support DMs in the S->M and M->S directions.


src/BridgedRoom.ts
```diff
@@ -206,10 +224,30 @@ export class BridgedRoom {
...
 
         this.dirty = true;
     }
 
+    public updateUsingChannelInfo(channelInfo: ConversationsInfoResponse) {
+        const chan = channelInfo.channel;
+        this.setValue("isPrivate", chan.is_private);
+        if (chan.is_channel) {
+            this.setValue("slackType", "channel");
+        } else if (chan.is_group) {
```
auscompgeek: Do we want to differentiate mpims and regular groups here?

src/OAuth2.ts
```diff
@@ -37,6 +37,13 @@ const REQUIRED_SCOPES = [
     "bot",
 ];
 
+const PUPPET_SCOPES = [ // See https://stackoverflow.com/a/28234443
+    "identify",
+    "read",
+    "post",
```
auscompgeek: From what I can tell, the `client` scope grants basically everything - wee-slack and I believe emacs-slack both only request client. Do we actually need these scopes?

Half-Shot: I guess not :)

src/BridgedRoom.ts
```diff
@@ -206,10 +224,30 @@ export class BridgedRoom {
...
+            this.setValue("slackType", "group");
+        } else if (chan.is_im) {
```
turt2live: we can probably give `is_im` a better name

src/Main.ts
```diff
@@ -663,6 +635,72 @@ export class Main {
...
+                body: "This slack bridge instance doesn't support private messaging.",
+                msgtype: "m.notice",
+            });
+            await intent.leave();
+            return;
+        }
+
+        const slackGhost = await this.getExistingSlackGhost(recipient);
+        if (!slackGhost || !slackGhost.teamId) {
+            // TODO: Create users dynamically who have never spoken.
```
turt2live: please raise an issue for this so we can track it

src/Main.ts
```diff
@@ -737,6 +776,21 @@ export class Main {
...
+            this.incRemoteCallCounter(method);
+        });
+        let puppetsWaiting: Promise<unknown> = Promise.resolve();
+        if (this.slackRtm) {
+            const puppetEntries = await this.datastore.getPuppetedUsers();
+            puppetsWaiting = Promise.all(puppetEntries.map(async (entry) => {
+                try {
+                    return this.slackRtm!.startUserClient(entry);
+                } catch (ex) {
+                    log.info("Failed to start puppet client");
```
turt2live: info? should probably be warn or error, also with the exception.

src/datastore/NedbDatastore.ts
```diff
@@ -149,4 +149,29 @@ export class NedbDatastore implements Datastore {
             });
         });
     }
+
+    public async setPuppetToken(): Promise<void> {
+        // Puppeting not supported by NeDB yet - noop
```
turt2live:
```suggestion
        // Puppeting not supported by NeDB - noop
```

src/datastore/postgres/PgDatastore.ts
```diff
@@ -190,6 +190,51 @@ export class PgDatastore implements Datastore {
         } as TeamEntry;
     }
 
+    public async setPuppetToken(teamId: string, slackUser: string, matrixId: string, token: string): Promise<void> {
+        await this.postgresDb.none("INSERT INTO puppets VALUES (${slackUser}, ${teamId}, ${matrixId}, ${token})" +
+                                        "ON CONFLICT ON CONSTRAINT cons_puppets_uniq DO UPDATE SET token = ${token}", {
```
turt2live: the indentation on this queries and the ones below are all over the place

src/BridgedRoom.ts
Half-Shot: This is a slack response, so this func is just handling it.

src/BridgedRoom.ts
```diff
@@ -206,10 +224,32 @@ export class BridgedRoom {
...
     }
 
+    public updateUsingChannelInfo(channelInfo: ConversationsInfoResponse) {
+        const chan = channelInfo.channel;
+        this.setValue("isPrivate", chan.is_private);
+        if (chan.is_channel) {
+            this.setValue("slackType", "channel");
+        } else if (chan.is_group) {
+            this.setValue("slackType", "group");
+        } else if (chan.is_mpim) {
```
auscompgeek: looking at this, I think this is the wrong way around. `is_group` seems to be true whenever `is_mpim` is true

PR #202, opened 2019-09-05
Update config sample and schema

Half-Shot:
We've made a few changes and not bothered to update the sample config & schema. Let's do that.


config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+  logging: "silent" # Logging level specific to RTM traffic.
+
+# Allow users to add channels dynamically by using oauth, or puppet themselves.
+# Optional
+oauth2:
+  client_id: ""
+  client_secret: ""
+  # A prefix similar to inbound_uri_prefix for oauth2 requests. inbound_uri_prefix will be used if this is not set
+  # Optional
+  redirect_prefix: ""
```
Cadair: Add an example here incase I am not using `inbound_uri_prefix`?

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
-      - "./error.log": "error"
+        "./error.log": "error"
 
-# Optional
-dbdir: /path/to/data/storage/
+# Settings for database connections.
+# Optional if using NeDB.
+db:
+  # Which database engine to use.
+  engine: "postgres" # Only postgres is supported at the moment.
```
Cadair: is a valid engine `neb`? you should document both options here (especially the not deprecated one that maintains behaviour).

src/app.ts
```diff
@@ -29,7 +29,7 @@ const cli = new Cli({
         reg.setId(AppServiceRegistration.generateToken());
         reg.setHomeserverToken(AppServiceRegistration.generateToken());
         reg.setAppServiceToken(AppServiceRegistration.generateToken());
-        reg.setSenderLocalpart(config.bot_username);
+        reg.setSenderLocalpart("slackbot");
```
Cadair: Surely this wants to be configurable?

Half-Shot: These are just the defaults, you are expected to go configure them yourself.

config/config.sample.yaml
Half-Shot: `nedb` could be, though at the moment we just assume it's in use if you don't specify db. I guess it can't hurt though.

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+# Optional
 logging:
-  console: "info"
+  # Log level for console output
+  console: "info" # One of "debug", "info", "warning", "error", "silent"
+   # Files that the bridge may log into. The key is the minimum level to report.
   files:
       - "./debug.log": "info"
-      - "./error.log": "error"
+        "./error.log": "error"
```
turt2live: why is `files` even a list?

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+# Optional
+tls:
+  key_file: /path/to/tls.key
+  crt_file: /path/to/tls.crt
+
+# Optional if slack_hook_port and inbound_uri_prefix are defined, required otherwise.
+rtm:
+  # Use the RTM API to listen for requests, which does not require
+  # the bridge to listen on a port. This should be on unless you have
+  # specific needs.
```
turt2live: a description of at least one of those use cases would be appreciated

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+  client_secret: ""
+  # A prefix similar to inbound_uri_prefix for oauth2 requests. inbound_uri_prefix will be used if this is not set
+  # Optional
+  redirect_prefix: ""
+
+# The prefix to give slack users 
+username_prefix: "slack_" # Required
+
+homeserver: # Required
+  # The server_name of your homeserver
```
turt2live: Instead of saying the field name again, can we say something like `The domain of your homeserver. Eg: "matrix.org"`

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
   server_name: my.server
+  # The URL which the bridge can reach the homeserver on.
+  url: http://localhost:8008
+  # The URL which slack can reach media on.
```
turt2live: Is this required? Should we just say "The public-facing URL for your homeserver"?

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+  media_url: "http://my.server.here"
 
+# Optional
 logging:
-  console: "info"
+  # Log level for console output
```
turt2live:
```suggestion
  # The default log level for the bridge
```

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+   # Files that the bridge may log into. The key is the minimum level to report.
   files:
       - "./debug.log": "info"
-      - "./error.log": "error"
+        "./error.log": "error"
 
-# Optional
-dbdir: /path/to/data/storage/
+# Settings for database connections.
+# Optional if using NeDB.
```
turt2live: how do I know if I'm using NeDB?

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
+db:
+  # Which database engine to use.
+  engine: "postgres" # Only postgres is supported at the moment.
+  connectionString: "postgresql://slackbridge_user:pass@localhost/slack_bridge" # A postgres connection string
```
turt2live:
```suggestion
  connectionString: "postgresql://slackbridge_user:pass@localhost/slack_bridge?sslmode=require" # A postgres connection string
```

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
 
+# Specify the location of NeDB datastore files
+# This option is DEPRECATED in favor of using db. See docs/datastore.md
+dbdir: "."
+
+# Use a matrix room to issue commands to the bridge
 # Optional
 matrix_admin_room: "!aBcDeF:matrix.org"
 
+# Enable metrics reporting on http://0.0.0.0:bridgePort/metrics which can be scraped by prometheus
```
turt2live: Metrics should always be optionally on a different port. Always.

config/config.sample.yaml
```diff
@@ -1,25 +1,67 @@
...
-dbdir: /path/to/data/storage/
+# Settings for database connections.
+# Optional if using NeDB.
+db:
+  # Which database engine to use.
+  engine: "postgres" # Only postgres is supported at the moment.
+  connectionString: "postgresql://slackbridge_user:pass@localhost/slack_bridge" # A postgres connection string
 
+# Specify the location of NeDB datastore files
+# This option is DEPRECATED in favor of using db. See docs/datastore.md
```
turt2live:
```suggestion
# This option is DEPRECATED in favor of using the database options above. See docs/datastore.md for more information
```

config/config.sample.yaml
Half-Shot: files is an object?

config/config.sample.yaml
Half-Shot: Huh, but it's not the default? It's the actual level at which logs will be outputted to the console/stdout?

config/config.sample.yaml
turt2live: not according to line 13:47

config/config.sample.yaml
Half-Shot: Sure, but not in this PR. Supporting this is non-trivial.

config/config.sample.yaml
turt2live: Which would be everything that's unhandled, ergo default. The comment doesn't add anything otherwise, so I would recommend changing it or removing it.

config/config.sample.yaml
Half-Shot: Will be explained in the datastore.md doc.

config/config.sample.yaml
Half-Shot: argh

config/config.sample.yaml
Half-Shot: Probably better to remove it.

config/config.sample.yaml
Half-Shot: https://github.com/matrix-org/matrix-appservice-slack/issues/209

PR #208, opened 2019-09-06
Npm Audit

Half-Shot:
Since we're gearing up for a release anyway.


Half-Shot:
I'm not sure why I bothered requiring a review for this one.

PR #220, opened 2019-09-09
Do not revoke tokens

Half-Shot:
Revoking tokens currently breaks the bridge, so do not do it.


Half-Shot:
>we should unbreak the bridge rather than disable logout.

I was probably unclear. This is a Slack problem where revoking your *personal* OAuth token will also revoke the *bot* token which will break all rooms on the workspace that are bridged across. There isn't an alternative API so we are a bit stuck.

Cadair:
This is only a problem if it's the user who provisioned the bridge right? If it's another user then it's ok?

Half-Shot:
>This is only a problem if it's the user who provisioned the bridge right? If it's another user then it's ok?

Good question, I really need to test this.

PR #225, opened 2019-09-12
Move room storage into a dedicated class

Half-Shot:
This is done in part to reduce dependence on Main, and for readability. A full set of tests are included.


src/AdminCommands.ts
```diff
@@ -115,16 +115,14 @@ export class AdminCommands {
         return new AdminCommand(
             "show",
             "show a single connected room",
-            ({respond, channel, channel_id, room}) => {
+            ({respond, channel_id, room}) => {
```
benparsons: Spoke IRL, HS to modify

Half-Shot:
(I think @benparsons was otherwise happy IRL)

PR #229, opened 2019-09-13
Migration script fixes for team refactoring

Half-Shot:
This tweaks the migration script so it *should* migrate the old format stores into the new format with separated team tokens from rooms and all that fancy stuff.

This script has been tested to ensure it works with the 0.3.2 data format.

### DIFF
--- a/src/SlackClientFactory.ts
+++ b/src/SlackClientFactory.ts
@@ -1,5 +1,5 @@
 import { Datastore, TeamEntry } from "./datastore/Models";
-import { WebClient, WebClientOptions } from "@slack/web-api";
+import { WebClient, WebClientOptions, LogLevel } from "@slack/web-api";
 import { Logging } from "matrix-appservice-bridge";
 import { TeamInfoResponse, AuthTestResponse, BotsInfoResponse, UsersInfoResponse } from "./SlackResponses";
 
@@ -158,6 +158,7 @@ export class SlackClientFactory {
                 info: webLog.info.bind(webLog),
                 error: webLog.error.bind(webLog),
             },
+            logLevel: LogLevel.DEBUG,
             ...opts,
         });
         try {
--- a/src/SlackClientFactory.ts
+++ b/src/SlackClientFactory.ts
@@ -18,7 +18,7 @@ interface RequiredConfigOptions {
 export class SlackClientFactory {
     private teamClients: Map<string, WebClient> = new Map();
     private puppets: Map<string, WebClient> = new Map();
-    constructor(private datastore: Datastore, private config: RequiredConfigOptions, private onRemoteCall: (method: string) => void) {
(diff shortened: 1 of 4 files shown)


src/SlackClientFactory.ts
```diff
@@ -140,14 +140,16 @@ export class SlackClientFactory {
...
-        const opts = this.config.slack_client_opts;
+        const opts = this.config ? this.config.slack_client_opts : undefined;
         const slackClient = new WebClient(token, {
             logger: {
                 setLevel: () => {}, // We don't care about these.
                 setName: () => {},
                 debug: (msg: string) => {
                     // non-ideal way to detect calls to slack.
-                    const match = /apiCall\('([\w\.]+)'\) start/.exec(msg);
+                    webLog.debug.bind(webLog);
```
turt2live: leftover by accident?

src/scripts/migrateToPostgres.ts
```diff
@@ -67,62 +73,131 @@ async function main() {
...
+
+    const preTeamMigrations = () => Promise.all(allRooms.map(async (room, i) => {
+        // This is an old format remote
+        // tslint:disable-next-line: no-any
+        const at = (room.remote as any).access_token;
+        if (!at) {
+            return;
+        }
+        try {
+            if (at) {
```
turt2live: usually the if comes before the try unless the if has multiple branches

src/scripts/migrateToPostgres.ts
```diff
@@ -67,62 +73,131 @@ async function main() {
...
 
-    const slackUserMigrations = allSlackUsers.map(async (user, i) => {
+    const slackUserMigrations = () => Promise.all(allSlackUsers.map(async (user, i) => {
         // tslint:disable-next-line: no-any
-        const ghost = SlackGhost.fromEntry(null as any, user, null);
-        await pgres.upsertUser(ghost);
+        let ghost = SlackGhost.fromEntry(null as any, user, null);
+        if (!ghost.slackId || !ghost.teamId) {
+            const localpart = ghost.userId.split(":")[0];
+            // TODO: we are making an assumption here that the prefix ends with _
```
turt2live:
```suggestion
            // XXX: we are making an assumption here that the prefix ends with _
```

src/scripts/migrateToPostgres.ts
```diff
@@ -67,62 +73,131 @@ async function main() {
...
-        );
-        log.info("Completed migration");
-    } catch (ex) {
-        log.error("An error occured while migrating databases:");
-        log.error(ex);
-        log.error("Your existing databases have not been modified, but you may need to drop the postgres table and start over");
-    }
+    }));
+    await eventMigrations();
+    log.info("Finished eventMigrations");
```
turt2live: Would be good to pair these with "Starting eventMigrations" and such, to show progress.

src/SlackClientFactory.ts
Half-Shot: Nope, intentional.

PR #231, opened 2019-09-16
Add snapcraft

devec0:
Adds necessary wrapper and snapcraft metadata to build and deploy this service as a snap.

If/when this is merged, it will need to be deployed to the snapcraft store. This can be done automatically using snapcraft's GitHub integration or via a CI system. I can also assist to publish the builds if desired. This will allow folks to very easily install this bridge and keep it up to date as new releases are made.

Signed-off-by: James Hebden <james@ec0.io>

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,60 @@
+name: matrix-appservice-slack
+base: core18
+version: git 
+summary: Matrix appservice bridge for Slack rooms
+description: |
+  This bridge allows you to connect Slack channels to Matrix rooms.
+
+grade: stable 
+confinement: strict 
+
+apps:
+  matrix-appservice-slack: 
+    command: 'bin/matrix-appservice-slack-wrapper'
+    plugs: [network-bind, network]
+    daemon: simple
+parts:
+  nodejs-lts:
+    plugin: nil
+    override-build: |
+      NODE_VERSION=12.13.0
+      case $SNAPCRAFT_ARCH_TRIPLET in
+      x86_64-*)
+        NODE_ARCH="x64"
+        ;;
+      aarch64-*)
+        NODE_ARCH="arm64"
+        ;;
+      arm-linux-gnueabihf)
+        NODE_ARCH="armv7l"
+        ;;
+      powerpc64le-*)
+        NODE_ARCH="ppc64le"
+        ;;
+      s390x-*)
+        NODE_ARCH="s390x"
+        ;;
+      esac
+      echo "Fetching node $NODE_VERSION for $NODE_ARCH"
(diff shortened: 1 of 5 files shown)


devec0:
Thanks for the quick response and guidance, I'd managed to miss those guidelines entirely. I've updated the commit and PR, and rebased off of the develop branch. Let me know if further changes are required.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,60 @@
...
+apps:
+  matrix-appservice-slack: 
+    command: 'bin/matrix-appservice-slack-wrapper'
+    plugs: [network-bind, network]
+    daemon: simple
+parts:
+  nodejs-lts:
+    plugin: nil
+    override-build: |
+      NODE_VERSION=10.16.3
```
Half-Shot: Use 12.X if possible.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,60 @@
+name: matrix-appservice-slack
+base: core18
+version: git 
+summary: Node.js Matrix bridge for Slack
+description: |
+  A bridge between Matrix and Slack. Currently the bridge is in Beta.
```
Half-Shot: This is now out of date. Use the line from the README

snap/snapcraft.yaml
```diff
@@ -0,0 +1,60 @@
+name: matrix-appservice-slack
+base: core18
+version: git
```
Half-Shot: What does this mean?

snap/snapcraft.yaml
devec0: No problem, I'll switch this to the latest v12 LTS

snap/snapcraft.yaml
devec0: OK, revised commit incoming.

snap/snapcraft.yaml
devec0: `version:  git` tells snapcraft to use the current git commit as the basis for the snap version details, when building a snap. This version then becomes available for pushing to the store and releasing via one of the channels, beta, edge or stable.

The git versioning support in snapcraft detects tags and branches sharing the current commit ID and is able to build regular semvers using that data.

Essentially, this just means that nobody needs to manually set the version on each release, so long as tags are in place which `git describe` will produce a sensible version string for.

devec0:
I have added documentation about building and installing the 'devmode' snap, as until this package is published on the snapcraft store, a simple `snap install matrix-appservice-slack` will not work, but a local install with `--devmode` will work fine.

I've made sure the commit and PR are signed off as well, and added a changelog entry in addition to addressing your review comments above.

Let me know if anything further is required.

PR #235, opened 2019-09-17
Support puppeted reactions/redactions

Half-Shot:


src/BridgedRoom.ts
```diff
@@ -254,7 +254,8 @@ export class BridgedRoom {
 
         // TODO: This only works once from matrix as we are sending the event as the
```
turt2live: irrelevant now?

Cadair: Still relevant for all non-puppeted users.

PR #236, opened 2019-09-18
Remove user token reference

Half-Shot:


Cadair:
fook, this is going to break all of my picard deployments...

Half-Shot:
>fook, this is going to break all of my picard deployments...

We don't use it anymore anyway :cry:

Cadair:
> > fook, this is going to break all of my picard deployments...
> 
> We don't use it anymore anyway cry

yeah I know.

PR #237, opened 2019-09-18
Deduplicate puppeted messages

### LINKED ISSUE #234: Puppets echo back across the bridge

Half-Shot:
Fixes #234

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -48,6 +48,7 @@ interface ISlackChatMessagePayload extends ISlackToMatrixResult {
 }
 
 const RECENT_MESSAGE_MAX = 10;
+const PUPPET_INCOMING_DELAY_MS = 5000;
 
 export class BridgedRoom {
     public get isDirty() {
@@ -431,6 +432,11 @@ export class BridgedRoom {
     }
 
     public async onSlackMessage(message: ISlackMessageEvent, content?: Buffer) {
+        if (this.slackTeamId && message.user) {
+            // This just checks if the user *could* be puppeted. If they are, delay handling their incoming messages.
+            const hasPuppet = null !== await this.main.datastore.getPuppetTokenBySlackId(this.slackTeamId, message.user);
+            await new Promise((resolve) => setTimeout(resolve, PUPPET_INCOMING_DELAY_MS));
+        }
         if (this.recentSlackMessages.includes(message.ts)) {
             // We sent this, ignore.
             return;
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -254,7 +254,8 @@ export class BridgedRoom {
 
         // TODO: This only works once from matrix as we are sending the event as the
         // bot user.
-        const res = await this.botClient.reactions.add({
+        const client = (await this.main.clientFactory.getClientForUser(this.SlackTeamId!, message.sender)) || this.botClient;
+        const res = await client.reactions.add({
             as_user: false,
             channel: this.slackChannelId,
             name: emojiKeyName,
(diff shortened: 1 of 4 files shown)


src/BridgedRoom.ts
```diff
@@ -429,6 +450,10 @@ export class BridgedRoom {
         if (message.user_id === this.team!.user_id) {
             return;
         }
+        if (this.recentSlackMessages.includes(`reactadd:${message.reaction}:${message.user_id}`)) {
+            // We sent this, ignore.
```
Cadair: We send this, add it to the events db, and I think you need the parent event ts for that, even if you don't need it to make the reaction unique.

src/BridgedRoom.ts
```diff
@@ -251,16 +255,27 @@ export class BridgedRoom {
...
         // bot user.
-        const res = await this.botClient.reactions.add({
+        const res = await client.reactions.add({
             as_user: false,
             channel: this.slackChannelId,
             name: emojiKeyName,
             timestamp: event.slackTs,
         });
 
+        if (id) {
```
Cadair: Don't we also want to put the event ids of reactions into the db for the bot?

src/BridgedRoom.ts
```diff
@@ -429,6 +460,10 @@ export class BridgedRoom {
         if (message.user_id === this.team!.user_id) {
```
Cadair: I think we want to drop this clause, because we need to handle events that the bot sent to add them to the db.

src/BridgedRoom.ts
```diff
@@ -47,6 +47,9 @@ interface ISlackChatMessagePayload extends ISlackToMatrixResult {
     icon_url?: string;
 }
 
+const RECENT_MESSAGE_MAX = 10;
```
Cadair: 10 seems on the low side? It's not like even 100 would use much ram?

src/BridgedRoom.ts
```diff
@@ -47,6 +47,9 @@ interface ISlackChatMessagePayload extends ISlackToMatrixResult {
...
+const PUPPET_INCOMING_DELAY_MS = 1500;
```
Cadair: This seems high? I suppose it's only for messages a logged in user sends from slack?

src/BridgedRoom.ts
Half-Shot: Leaving the DB stuff to you :stuck_out_tongue:

src/BridgedRoom.ts
Half-Shot: The bot's messages are automatically filtered out

src/BridgedRoom.ts
Half-Shot: Can we leave this for your PR?

src/BridgedRoom.ts
Half-Shot: My worry is many active slack rooms would start to chug a bit, although 10 is a bit low. How about 25?

src/BridgedRoom.ts
Half-Shot: yep.

src/BridgedRoom.ts
Cadair: yep

src/BridgedRoom.ts
Cadair: What's going to slow it down? the `thing in list` operations? That doesn't seem like something that should be slow moving?

Half-Shot: I'm thinking of matrix.org and it's thousands upon thousands of rooms :(

Half-Shot:
Merged in 3fb4637

PR #238, opened 2019-09-18
Do not send messages M->S if we have no content to send

Half-Shot:
The bridge will attempt to send messages that we might not even understand. This should be filtered out before we hit Slack's API.

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -345,6 +345,11 @@ export class BridgedRoom {
             username: user.getDisplaynameForRoom(message.room_id) || matrixToSlackResult.username,
         };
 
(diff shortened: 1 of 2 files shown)


src/BridgedRoom.ts
```diff
@@ -345,6 +345,11 @@ export class BridgedRoom {
             username: user.getDisplaynameForRoom(message.room_id) || matrixToSlackResult.username,
         };
 
+        if (!body.attachments && !body.text) {
```
Cadair: What is attachments?!

Half-Shot: dunno, something we use in substitutions i guess :p

PR #246, opened 2019-09-23
Add more logging to startup routine

Half-Shot:
Deploying this to matrix.org failed, but we lack a lot of logging to understand what went wrong. This should hopefully improve things.

### DIFF
--- a/src/SlackClientFactory.ts
+++ b/src/SlackClientFactory.ts
@@ -156,7 +156,7 @@ export class SlackClientFactory {
                     // non-ideal way to detect calls to slack.
                     webLog.debug.bind(webLog);
                     if (!this.onRemoteCall) { return; }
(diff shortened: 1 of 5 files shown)


src/SlackClientFactory.ts
```diff
@@ -156,7 +156,7 @@ export class SlackClientFactory {
                     // non-ideal way to detect calls to slack.
                     webLog.debug.bind(webLog);
                     if (!this.onRemoteCall) { return; }
-                    const match = /apiCall\('([\w\.]+)'\) start/.exec(msg[0]);
+                    const match = /apiCall\('([\w\.]+)'\) start/.exec(msg);
```
benparsons: what's this change?

Half-Shot: Message is a string, we were regexing against the first char

Half-Shot: Message is a string, we were regexing against the first char

PR #253, opened 2019-09-24
Fix multi-person DMs being registered as groups

auscompgeek:
`is_group` is set in addition to `is_mpim` by the Slack API.

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -183,10 +183,11 @@ export class BridgedRoom {
         this.setValue("isPrivate", chan.is_private);
         if (chan.is_channel) {
             this.setValue("slackType", "channel");
+        } else if (chan.is_mpim) {
+            // note: is_group is also set for mpims, so order is important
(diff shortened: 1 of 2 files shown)


src/BridgedRoom.ts
```diff
@@ -183,10 +183,10 @@ export class BridgedRoom {
         this.setValue("isPrivate", chan.is_private);
         if (chan.is_channel) {
             this.setValue("slackType", "channel");
-        } else if (chan.is_group) {
-            this.setValue("slackType", "group");
         } else if (chan.is_mpim) {
             this.setValue("slackType", "mpim");
+        } else if (chan.is_group) {
```
Cadair: Can you add a comment here so we remember the ordering is important?

auscompgeek: I've put the comment in the above if statement instead, since I think it makes more sense there.

Half-Shot:
This doesn't look dangerous, by the powers invested in me I will merge this.

PR #257, opened 2019-09-24
If a message could not be deleted, drop the request

Half-Shot:


changelog.d/257.bugfix
```diff
@@ -0,0 +1 @@
+Fix .toUpperCase() errors
```
auscompgeek: uh.

Half-Shot: This is actually the user facing error, so it's fine to put this into the changelog (although I should probably have made this more detailed).

PR #261, opened 2019-09-25
Always send a 200 on an Event API request

Half-Shot:
This PR refactors our event handling so that we correctly handle Event API requests, even if using the RTM API too. This bug came up when matrix.org switched on RTM and stopped handling Events for nearly every team. While this means that RTM has more or less taken over the matrix.org deployment (yay), it also made slack turn off our Events API subscription (boo).


ara4n:
lgtm - and yay for UTing :)

PR #262, opened 2019-09-27
Ignore tombstone events

Half-Shot:
Unless we are dealing with message_deleted


src/SlackEventHandler.ts
```diff
@@ -160,6 +161,11 @@ export class SlackEventHandler extends BaseSlackHandler {
             return;
         }
 
+        if (event.subtype !== "message_deleted" && event.message && event.message.subtype === "tombstone") {
+            throw Error("ignored");
+            // Filter out tombstones early, we only care about them on deletion.
+            return;
```
tulir: Return after throw? :thinking:

PR #264, opened 2019-09-27
Limit upload size

Half-Shot:
This fixes an issue where uploading large files will crash the bridge. The fix here is to provide a URL as a fallback if the file exceeds the configured limit.

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -611,7 +611,7 @@ export class BridgedRoom {
             };
             return ghost.sendMessage(this.MatrixRoomId, matrixContent, channelId, eventTS);
         } else if (message.files) { // A message without a subtype can contain files.
-            const uploadSize = this.main.config.homeserver.max_upload_size;
+            const maxUploadSize = this.main.config.homeserver.max_upload_size;
             for (const file of message.files) {
                 if (!file.url_private) {
                     // Cannot do anything with this.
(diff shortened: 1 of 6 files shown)


auscompgeek:
Would it make sense in this scenario to send the "caption" in the same message if there is one?

Half-Shot:
> Would it make sense in this scenario to send the "caption" in the same message if there is one?

I decided against it for consistency, and because it's possible for other files in the set to be small enough to pass the upload limit. You could do it that way, but it would be a logical mess to follow in the code :/

src/BridgedRoom.ts
```diff
@@ -611,11 +611,13 @@ export class BridgedRoom {
             };
             return ghost.sendMessage(this.MatrixRoomId, matrixContent, channelId, eventTS);
         } else if (message.files) { // A message without a subtype can contain files.
+            const uploadSize = this.main.config.homeserver.max_upload_size;
```
benparsons: Can we make it `maxUploadSize`? Makes the check line clearer (`file.size > uploadSize`)

src/BridgedRoom.ts
```diff
@@ -648,13 +650,21 @@ export class BridgedRoom {
...
+                        if (link) {
+                            log.info("File too large, sending as a link");
+                            const messageContent = {
+                                body: `${link} (${file.name})`,
+                                format: "org.matrix.custom.html",
+                                formatted_body: `<a href="${link}">${file.name}</a>`,
+                                msgtype: "m.text",
+                            };
+                            await ghost.sendMessage(this.matrixRoomId, messageContent, channelId, eventTS);
+                        }
```
benparsons: Is it ever possible for `link` to not exist? Is there a need for an else condition?

PR #266, opened 2019-09-30
Set unfurl_links for Slack messages we post

auscompgeek:


Half-Shot:
Thanks, this looks good but we're going to leave this for the next release.

PR #269, opened 2019-09-30
Fix webhooks not being able to send messages to Matrix

Half-Shot:
We assume here that teamId is set, where it may not be since webhooks do not have a teamId. This unbreaks bridging slack messages into Matrix by only calling toUpperCase inside the SlackGhost, if the values are given.


Half-Shot:
Thanks folks.

PR #280, opened 2019-10-04
Refactor Matrix -> Slack subsitutions

### LINKED ISSUE #183: Substrings of pills should not turn into Slack mentions

Half-Shot:
Hopefully fixes #183

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -571,7 +571,6 @@ export class Main {
                         endTimer({outcome: "fail"});
                         return;
                     }
-                    return;
                 }
             }
             try {
@@ -581,7 +580,6 @@ export class Main {
                 endTimer({outcome: "fail"});
                 return;
             }
-            return;
         }
 
         endTimer({outcome: success ? "success" : "dropped"});
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -552,6 +552,8 @@ export class Main {
             endTimer({outcome: "success"});
         }
 
+        let success = false;
+
         // Handle a m.room.message event
         if (ev.type === "m.room.message" || ev.content) {
             if (ev.content["m.relates_to"] !== undefined) {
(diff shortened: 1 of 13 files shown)


src/Main.ts
```diff
@@ -552,33 +556,35 @@ export class Main {
...
                     try {
-                        await room.onMatrixEdit(ev);
+                        success = await room.onMatrixEdit(ev);
                     } catch (e) {
                         log.error("Failed processing matrix edit: ", e);
                         endTimer({outcome: "fail"});
                         return;
                     }
-                    endTimer({outcome: "success"});
                     return;
```
auscompgeek: Don't you still want to `endTimer` here?

src/tests/unit/substitutionsTest.ts
```diff
@@ -26,29 +29,219 @@ const displayMap = {
...
+                    body: "@room Hello everyone!",
+                },
+                sender: "@alice:localhost",
+            }, fakeMain, "");
+            expect(res).to.deep.equal({
+                link_names: true,
+                text: "@channel Hello everyone!",
+                username: "@alice:localhost",
+            });
+        });
```
auscompgeek: This looks like the same test twice?

Half-Shot: Copy paste bug, good catch.

src/substitutions.ts
```diff
@@ -341,11 +326,16 @@ export function replacementFromDisplayMap(str: string, displaymap: IDisplayMap):
...
  */
-async function plainTextSlackMentions(main: Main, s: string, roomId: string) {
-    return replacementFromDisplayMap(
-        s,
-        await substitutions.getDisplayMap(main, roomId),
-    );
+async function plainTextSlackMentions(main: Main, body: string, teamId: string) {
+    const users = await main.datastore.getAllUsersForTeam(teamId);
+    for (const user of users) {
+        const displayName = `@${user.display_name}`;
```
Cadair: So you went with `@` prefix?

src/substitutions.ts
```diff
@@ -34,37 +39,22 @@ interface IDisplayMap {
...
 }
 
-interface IFirstWordMap {
-    [firstword: string]: [IDisplayMap];
+interface PillItem {
+    id: string;
+    text: string;
 }
 
 export interface ISlackToMatrixResult {
```
Cadair: This interface feels like it's named backwards? It's only used in `matrixToSlack`?

src/tests/unit/substitutionsTest.ts
```diff
@@ -26,29 +29,219 @@ const displayMap = {
...
+                },
+                sender: "@alice:localhost",
+            }, fakeMain, "");
+            expect(res).to.deep.equal({
+                link_names: true,
+                text: "Let's talk about &lt;html&gt; and the &amp; character",
+                username: "@alice:localhost",
+            });
+        });
+        it ("should replace @channel with @room", async () => {
```
Cadair: Also `@here` and `@everyone`?

Half-Shot: The title is wrong, should be replace @\room with @\channel. And Matrix only supports @\room

src/substitutions.ts
Half-Shot: It is :(

auscompgeek:
How will this behave if there are two hypothetical users with the display names `Alice` and `Alice Bob` in the same channel, for example?

Half-Shot:
> How will this behave if there are two hypothetical users with the display names `Alice` and `Alice Bob` in the same channel, for example?

Good catch as always, I've provisioned for this now too.

Issue #265, opened 2019-09-28
Optionally do not keep BridgedRoom in memoey


Half-Shot:
Aside from a memory improvement, this allows us to make changes in the datastore without having to reload the bridge. We could keep a cache for N seconds to avoid any performance hits.

PR #293, opened 2019-10-07
Membership syncing

Cadair:
This is a WIP to implement full membership syncing with slack. It will eventually support reacting to join and leave events on slack and then also creating ghosts for all the slack users in the room at startup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "postinstall": "npm run build",
     "start": "node ./lib/app.js",
-    "build": "tsc -sourcemap",
+    "build": "tsc",
     "test": "npm run test:unit && npm run test:integration",
     "test:unit": "mocha --reporter list --ui bdd --require ts-node/register --require source-map-support/register --recursive lib/tests/unit",
     "test:integration": "mocha --exit --reporter list --ui bdd --require ts-node/register --require source-map-support/register --recursive lib/tests/integration",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "postinstall": "npm run build",
     "start": "node ./lib/app.js",
-    "build": "tsc",
+    "build": "tsc -sourcemap",
     "test": "npm run test:unit && npm run test:integration",
     "test:unit": "mocha --reporter list --ui bdd --require ts-node/register --require source-map-support/register --recursive lib/tests/unit",
     "test:integration": "mocha --exit --reporter list --ui bdd --require ts-node/register --require source-map-support/register --recursive lib/tests/integration",
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -21,7 +21,7 @@ import { Main, METRIC_SENT_MESSAGES } from "./Main";
 import { default as substitutions, getFallbackForMissingEmoji, ISlackToMatrixResult } from "./substitutions";
 import * as emoji from "node-emoji";
 import { ISlackMessageEvent, ISlackEvent } from "./BaseSlackHandler";
-import { WebClient } from "@slack/web-api";
+import { WebClient, ConversationsMembersArguments } from "@slack/web-api";
 import { ChatUpdateResponse,
     ChatPostMessageResponse, ConversationsInfoResponse } from "./SlackResponses";
 import { RoomEntry, EventEntry, TeamEntry } from "./datastore/Models";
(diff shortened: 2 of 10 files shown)


src/BridgedRoom.ts
```diff
@@ -176,6 +176,40 @@ export class BridgedRoom {
...
+        }
+        const ghost = await this.main.getGhostForSlack(user_id, this.team.domain, this.team.id);
+        return ghost.intent.join(this.matrixRoomId);
+    }
+
+    private async getSlackChannelMembership() {
+        if (!this.botClient || !this.slackChannelId) {
+            return;
+        }
+        // TODO: Handle paging here
```
Half-Shot: It's almost certainly true that if botClient exists, we have a team. I would urge you to just return early if `this.teams` isn't defined. And then I would try store the bot user_id in a `const` at the start and use that const in the filter.

src/BridgedRoom.ts
```diff
@@ -176,6 +176,34 @@ export class BridgedRoom {
...
 
         this.dirty = true;
+
+        // This almost certainly isn't the best place for this
+        this.joinAllSlackUsers();
+    }
+
+    public async joinAllSlackUsers() {
+        const members = await this.getSlackChannelMembership();
+        if (!members) return;
```
Half-Shot:
```suggestion
```

Don't bother doing this, a `Promise.all([].map(...))` will do the right thing.

src/BridgedRoom.ts
```diff
@@ -176,6 +176,34 @@ export class BridgedRoom {
         this.puppetOwner = opts.puppet_owner;
 
         this.dirty = true;
+
+        // This almost certainly isn't the best place for this
+        this.joinAllSlackUsers();
```
Half-Shot: I would call `joinAllSlackUsers` during startup (queued, obviously), and in `actionLink`. Calling it in the constructor not only means we are not awaiting it, but also will cause unintended side effects by doing this in a constructor (e.g. if we do our cached rooms thing we talked about in #265)

src/BridgedRoom.ts
Cadair: The typescript compiler disagrees with you.

src/SlackEventHandler.ts
```diff
@@ -144,6 +151,26 @@ export class SlackEventHandler extends BaseSlackHandler {
         }
     }
 
+    protected async handleMemberJoined(event: ISlackEvent, teamId: string) {
```
Half-Shot: This should be `private`

src/SlackEventHandler.ts
```diff
@@ -144,6 +151,26 @@ export class SlackEventHandler extends BaseSlackHandler {
...
+        const teamDomain = (await this.main.getTeamDomainForMessage(undefined, teamId)).toLowerCase();
+        if ( !event.user) {
+            throw Error("No user given for event.");
+        }
+        const ghost = await this.main.getGhostForSlack(event.user, teamDomain, teamId);
+        const room = this.main.rooms.getBySlackChannelId(event.channel) as BridgedRoom;
+        await ghost.intent.join(room.MatrixRoomId);
+    }
+
+    protected async handleMemberParted(event: ISlackEvent, teamId: string) {
```
Half-Shot: This should be `private`

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "postinstall": "npm run build",
     "start": "node ./lib/app.js",
-    "build": "tsc",
+    "build": "tsc -sourcemap",
```
Half-Shot: put this in `tsconfig.json`?

src/BridgedRoom.ts
Half-Shot: Oh bother, because it can be undefined? Fair

src/BridgedRoom.ts
```diff
@@ -178,6 +178,32 @@ export class BridgedRoom {
...
+
+    private async joinUserToRoom(user_id) {
+        if (!this.team) return;
+        const ghost = await this.main.getGhostForSlack(user_id, this.team.domain, this.team.id);
+        return ghost.intent.join(this.matrixRoomId);
+    }
+
+    private async getSlackChannelMembership() {
+        if (!this.team || !this.botClient) return;
+        const our_user_id = this.team.user_id;
```
Half-Shot: `camelCase`, not `snake_cake`

src/Main.ts
```diff
@@ -274,13 +274,13 @@ export class Main {
         return `${(hs.media_url || hs.url)}/_matrix/media/r0/download/${mxcUrl.substring("mxc://".length)}`;
     }
 
-    public async getTeamDomainForMessage(message: any, teamId?: string) {
-        if (message.team_domain) {
+    public async getTeamDomainForMessage(message?: any, teamId?: string) {
```
Half-Shot: This feel like the wrong approach, why are we calling getTeamDomain on unknown messages?

src/Main.ts
```diff
@@ -796,16 +796,22 @@ export class Main {
         log.info(`Found ${entries.length} room entries in store`);
         const joinedRooms = await roomListPromise;
         i = 0;
-        await Promise.all(entries.map(async (entry) => {
+        const rooms = await Promise.all(entries.map(async (entry) => {
             i++;
             log.info(`[${i}/${entries.length}] Loading room entry ${entry.matrix_id}`);
             try {
-                await this.startupLoadRoomEntry(entry, joinedRooms, teamClients);
+                return await this.startupLoadRoomEntry(entry, joinedRooms, teamClients);
```
Half-Shot: return isn't needed

src/Main.ts
```diff
@@ -796,16 +796,22 @@ export class Main {
...
             } catch (ex) {
                 log.error(`Failed to load entry ${entry.matrix_id}, exception thrown`, ex);
             }
         }));
 
+        const roomMembersPromises = new PQueue({concurrency: STARTUP_TEAM_INIT_CONCURRENCY});
+        for (const room of rooms) {
+            if (!room) return;
+            roomMembersPromises.add(async () => room.joinAllSlackUsers());
```
Half-Shot: async isn't needed here, is it?

src/Main.ts
```diff
@@ -796,16 +796,22 @@ export class Main {
...
-                await this.startupLoadRoomEntry(entry, joinedRooms, teamClients);
+                return await this.startupLoadRoomEntry(entry, joinedRooms, teamClients);
             } catch (ex) {
                 log.error(`Failed to load entry ${entry.matrix_id}, exception thrown`, ex);
             }
         }));
 
+        const roomMembersPromises = new PQueue({concurrency: STARTUP_TEAM_INIT_CONCURRENCY});
+        for (const room of rooms) {
+            if (!room) return;
```
Half-Shot:
```suggestion
            if (!room) continue;
```

src/Main.ts
Cadair: I should probably rename the method, it's more that we want to get the domain from the teamId.

Half-Shot:
(Sorry, I may have done this in #332) Thanks for kicking us off though!

Cadair:
Did you add initial sync?

Half-Shot:
I did indeed.

Issue #292, opened 2019-10-07
Linking in messages not present in Slack


joenas:
Message:
```json
{
  "type": "m.room.message",
  "sender": "@joenas:gottsnack.net",
  "content": {
    "body": "[link](https://jonnev.se)",
    "format": "org.matrix.custom.html",
    "msgtype": "m.text",
    "formatted_body": "<a href=\"https://jonnev.se\">link</a>"
  },
}
```
<img width="437" alt="Screenshot 2019-10-07 at 11 26 07" src="https://user-images.githubusercontent.com/1104793/66300522-c0282a00-e8f5-11e9-90e1-af1b3a50fa81.png">
<img width="453" alt="Screenshot 2019-10-07 at 11 29 04" src="https://user-images.githubusercontent.com/1104793/66300523-c0282a00-e8f5-11e9-962a-116a340b6f36.png">

Half-Shot:
Will be fixed by #280

joenas:
Including the `m.notice` thing @Half-Shot ?

Half-Shot:
Nope, but that's a separate issue. We don't currently parse HTML from messages at all, so it won't be bridged. You should open a new issue for that though.

auscompgeek:
Just tested with the matrix.org instance. I now only see the URL, the link text is discarded.

Half-Shot:
Reference: https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22%3Chttp%3A%2F%2Fsomeurl%7Clike%20this%3E%22%7D

joenas:
It works the other way around so I solved it that way for now! Sending formatted Slack links converts properly in Matrix 👌🏻

jaller94:
Fixed by https://github.com/matrix-org/matrix-appservice-slack/pull/280

PR #299, opened 2019-10-12
Fix Markdown link substitution

auscompgeek:
Ref: #292


Half-Shot:
Ensure you fix the tests too :p

auscompgeek:
Oh yes, tests would be good wouldn't they.

Issue #304, opened 2019-10-21
Unclear doc for config


HarHarLinks:
I'm trying to set up the bridge on my a-dozen-people-homeserver using docker. The whole docker installation process is only so-so documented for docker-noobs to begin with. The issue I'm facing is generally that it's unclear what services are included in the container (e.g. postgresql? Seems not: ECONNREFUSED) and what means what. What features exactly are covered anyway and what of the optional settings do I need to set up? What will the OAuth allow me to do? Add channels on matrix or slack?

So I decided on nedb as it seems to be less of a hassle because: Where is postgresql? How to set it up?

I'm specifically facing this following resulting issue.

conf.yaml:
```
  engine: "nedb" # "postgres" or "nedb". If using "nedb", do not define connectionString
```
so I comment out the next line `connectionString`.

alas:
```
$ docker run -v ~synapse/bridges/slack/:/config/ matrixdotorg/matrix-appservice-slack

{"field":"data.db.connectionString","message":"is required","value":{"engine":"nedb"},"type":"object","schemaPath":["properties","db"]}
The field data.db.connectionString is [object Object] which is required
/usr/src/app/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:44
        throw e;
        ^

Error: Failed to validate file
    at ConfigValidator.validate (/usr/src/app/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:42:17)
    at Cli._loadConfig (/usr/src/app/node_modules/matrix-appservice-bridge/lib/components/cli.js:160:22)
    at Cli._assignConfigFile (/usr/src/app/node_modules/matrix-appservice-bridge/lib/components/cli.js:145:23)
    at Cli.run (/usr/src/app/node_modules/matrix-appservice-bridge/lib/components/cli.js:139:10)
    at Object.<anonymous> (/usr/src/app/lib/app.js:46:5)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10)
    at internal/main/run_main_module.js:17:11 {
  _validationErrors: [
    {
      field: 'data.db.connectionString',
      message: 'is required',
      value: [Object],
      type: 'object',
[...]

PR #305, opened 2019-10-22
Improve doc for docker installations

HarHarLinks:
Add some documentation for docker as well as general installations.

See also discussion in chat: [#slack_bridge:matrix.org](https://matrix.to/#/!vCpZvJFNjuiBkDAvEs:cadair.com/$Nfi-lEqlzld-ULaxzGUskyZSwzFEOhrnJIa6IjAu-3Q?via=half-shot.uk&via=matrix.org&via=sorunome.de)

See also #304

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -8,6 +8,7 @@ the `config.sample.yaml` file.
 
 These instructions assume you are using Synapse 1.4.0+. They should
 work for older releases and other homeservers, but configuration may vary.
+The bridge also requires a database, for more information see [datastores](datastores.md).
 
 ### From source
 
@@ -59,6 +60,9 @@ ever stuck, you can post a question in the
 3. Create a `config.yaml` file for global configuration. There is a sample
    one to begin with in `config/config.sample.yaml`. You should copy and
    edit as appropriate. The required and optional values are flagged in the config.
+   
+   If using docker, put this file in a new `config` directory which we will later
+   mount to the container.
 
 4. See [datastores](datastores.md) on how to setup a database with the bridge.
 
@@ -72,8 +76,8 @@ ever stuck, you can post a question in the
    or with docker:
    
 ```sh
-$ docker run -v /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \ 
-    -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack.yaml
+$ docker run --network=host -v /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \ 
+    -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack-registration.yaml
 ```
 
 6. Start the actual application service:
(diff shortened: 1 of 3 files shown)


HarHarLinks:
Signed-off-by: Kim Brose <kim.brose@rwth-aachen.de>

Half-Shot:
Thanks, this looks useful. Can you add a newsfile as [this doc](https://github.com/matrix-org/matrix-appservice-slack/blob/ea75b3d610822406500f9935f2acdc488aa6a53b/CONTRIBUTING.md#making-pull-requests)

auscompgeek:
Your newsfile needs to be a single line.

docs/getting_started.md
```diff
@@ -72,16 +76,21 @@ ever stuck, you can post a question in the
    or with docker:
    
 ```sh
-$ docker run -v /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \ 
-    -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack.yaml
+$ docker run --network=host -v /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \
```
Half-Shot: I'm worried about opening up the docker image to the hosts' network. I think I would rather suggest creating a seperate postgres contaienr and using `--link`?

src/Main.ts
```diff
@@ -886,7 +886,7 @@ export class Main {
                 teamId = await this.clientFactory.upsertTeamByToken(opts.slack_bot_token);
                 log.info(`Found ${teamId} for token`);
             } catch (ex) {
-                log.error("Failed to action link because the token couldn't used:", ex);
+                log.error("Failed to action link because the token couldn't be used:", ex);
```
Half-Shot: I'll let you off, but this should have been a seperate PR because it's not really relevant to the context :p

docs/getting_started.md
HarHarLinks: I agree, and I note and explain this later. But afaik the communication also needs to be able to reach the homeserver (which might be in another container, host, or other) and vice versa. I do not know of a stable way to route this, except outgoing via domain. Can you reserve/bind an IP to a container?

HarHarLinks: I have continued experimenting and created this docker-compose.yaml for setups running the bridge with docker, but postgres and homeserver natively.
It works for me right now, except for receiving messages from slack (however when i call the webhook url manually, it shows up in log).
This configuration can be started using `docker-compose up -d`.
```yaml
version: "2.4" # version 2 is needed to support setting gateway

services:

  slack:
    container_name: appservice-slack
    image: matrixdotorg/matrix-appservice-slack
    restart: always
    volumes:
      - "~synapse/bridges/slack/:/config/" # mount your config folder (where you keep config.yaml)
    ports:
      - "127.0.0.1:5858:5858" # must be reachable from homeserver, so variate this accordingly
      - "9898:9898" # must be reachable from the internet. remember to open your firewall
    networks:
      default:
        ipv4_address: 172.19.0.2 # bind to this ip to allow precise psql config and firewalling

# this network setup is for dockered appservice + native homeserver and db.
# if you docker these or other matrix services, you should consider joining them on the same virtual network.
networks:
  default:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: "172.19.0.0/16"
          gateway: 172.19.0.1
# note the absence of a dash on the gateway line
# https://github.com/docker/compose/issues/4456
```

In addition to this, postgres must be configured to listen and allow connections from docker.
edit postgresql.conf:
```conf
listen_addresses = 'localhost,172.19.0.1' # include docker slack: this is the gateway IP from docker-compose.yaml
# also check the configured port to use in config.yaml connectionString
```
append to pg_hba.conf:
```conf
# allow password authenticated connections from docker: this is the subnet from docker-compose.yaml
host    all             all             172.19.0.1/16           md5
```

Notable changes in appservice-slack config.yaml:
```yaml
homeserver:
  url: http://172.19.0.1:8008 # can also use the public facing IP and port, e.g. https://homeserver.domain:8448
db:
# connectionString: "postgresql://user:password@gateway:port/slack_bridge?sslmode=require" # A postgres connection string
# for example:
  connectionString: "postgresql://slackbridge_user:password@172.19.0.1:5433/slack_bridge?sslmode=require" # A postgres connection string
```

PR #313, opened 2019-11-03
Fix missing info, typo, and incorrect indentation

jeffcutsinger:
* The docker run command didn't expose any ports. Most configurations will probably want to do this, and I had to look in the Dockerfile to see what matrix-side port was.
* The docker config generation command output `slack.yaml`, not `slack-registration.yaml`, and I spent a little time trying to figure out if there were two files I needed (I thought maybe the initial run might generate a second file for the homeserver).
* The indentation was off for the block commands. This is admittedly an aesthetic choice, but this makes it a little clearer they are associated with their respective list item.
* Homserver -> Homeserver.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -64,24 +64,27 @@ ever stuck, you can post a question in the
 
 5. Generate the appservice registration file. This will be used by the
    Matrix homeserver. Here, you must specify the direct link the
-   **Matrix Homserver** can use to access the bridge, including the Matrix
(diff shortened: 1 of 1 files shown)


Half-Shot:
Thanks for the PR. Please read https://github.com/matrix-org/matrix-appservice-slack/blob/ea75b3d610822406500f9935f2acdc488aa6a53b/CONTRIBUTING.md and ensure that you have followed all the steps for contirbuting to the project.

PR #320, opened 2019-11-13
Allow a trailing colon followed by a space for mentions

Half-Shot:
Why a colon followed by a space, because Riot does mentions this way. Why do we care about the space? So we don't match the localparts of userIds.


Cadair:
Thanks Halfy

PR #323, opened 2019-11-18
Add support for syncing channel creation and removal dynamically

Half-Shot:
This PR adds support for creating and removing rooms as channels get created and removed from Slack, if configured to do so for a team. It will also, by default, sync across all public channels to the homeserver.


src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+        const { channel } = (await client.conversations.info({ channel: channelId })) as ConversationsInfoResponse;
+        if (!channel.is_channel || channel.is_private) {
+            log.warn("Not creating channel: Is either private or not a channel");
+            return;
+        }
+        const userId = (await this.main.datastore.getTeam(teamId))!.user_id;
+        const {user} = (await client.users.info({ user: userId })) as UsersInfoResponse;
+        await client.chat.postEphemeral({
+            user: creator,
+            text: `Hint: To bridge to Matrix, run the \`/invite @${user!.name}\` command in this channel.`,
```
Cadair: How about an option to use a user token and automatically invite the bot on the slack side?

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+            return;
+        }
+        return channelConfig.alias_prefix;
+    }
+
+    private async createRoomForChannel(teamId: string, creator: string, channel: ConversationsInfo): Promise<string> {
+        let intent;
+        let creatorUserId: string|undefined;
+        try {
+            creatorUserId = (await this.main.getGhostForSlack(creator, undefined, teamId)).userId;
```
Cadair: Do we really want all these rooms to be created by a ghost and have a ghost have PL 100?

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+        const alias = aliasPrefix ? `${aliasPrefix}${channel.name.toLowerCase()}` : undefined;
+        let topic: undefined|string;
+        if (channel.purpose) {
+            topic = channel.purpose.value;
+        }
+        log.debug("Creating new room for channel", channel.name, topic, alias);
+        const plUsers = {};
+        plUsers[this.main.botUserId] = 100;
+        if (creatorUserId) {
+            plUsers[creatorUserId] = 100;
```
Cadair: can we have the ability to add users to this list via config? I will want to add matrix side admins to all the rooms etc.

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+        const {user} = (await client.users.info({ user: userId })) as UsersInfoResponse;
+        await client.chat.postEphemeral({
+            user: creator,
+            text: `Hint: To bridge to Matrix, run the \`/invite @${user!.name}\` command in this channel.`,
+            channel: channelId,
+        });
+
+        // Create the room first.
+        try {
+            const roomId = await this.createRoomForChannel(teamId, creator, channel);
```
Cadair: can we also have a list of matrix users to invite to the new room in the config?

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+        let creatorUserId: string|undefined;
+        try {
+            creatorUserId = (await this.main.getGhostForSlack(creator, undefined, teamId)).userId;
+            intent = this.main.getIntent(creatorUserId);
+        } catch (ex) {
+            // Couldn't get the creator's mxid, using the bot.
+            intent = this.main.botIntent;
+        }
+        const aliasPrefix = this.getAliasPrefix(teamId);
+        const alias = aliasPrefix ? `${aliasPrefix}${channel.name.toLowerCase()}` : undefined;
```
Cadair: Also add a way of setting an avatar for the room via the config?

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+                matrix_room_id: roomId,
+                slack_channel_id: channelItem.id,
+                team_id: teamId,
+            });
+        } catch (ex) {
+            log.error("Failed to provision new room dynamically:", ex);
+        }
+    }
+
+    public async onChannelDeleted(teamId: string, channelId: string) {
```
Cadair: what about on slack channel archived?

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+        if (!channelItem.is_channel || channelItem.is_private) {
+            log.warn("Not creating channel: Is either private or not a channel");
+            return;
+        }
+        log.info(`Attempting to dynamically bridge ${channelItem.id} ${channelItem.name}`);
+        const userId = (await this.main.datastore.getTeam(teamId))!.user_id;
+        const {user} = (await client.users.info({ user: userId })) as UsersInfoResponse;
+        try {
+            await client.chat.postEphemeral({
```
Cadair: You shouldn't repeat this, you should have a method to handle it, because then you can have it also do different things.

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+                slack_channel_id: channelId,
+                team_id: teamId,
+            });
+        } catch (ex) {
+            log.error("Failed to provision new room dynamically:", ex);
+        }
+    }
+
+    private async syncChannel(teamId: string, channelItem: ConversationsInfo) {
+        if (!this.shouldTeamSync(teamId, "channel", channelItem.id)) {
```
Cadair: A lot of this method is repeated from the new room one above.

src/TeamSyncer.ts
```diff
@@ -0,0 +1,317 @@
...
+            log.info(`Not syncing`);
```
Cadair: This is not descriptive.

PR #326, opened 2019-11-19
Support user admin rooms

### LINKED ISSUE #205: User admin rooms
So users can configure the bridge for their own account without having to use an integration manager. This could be as easy as wrapping the provisioning commands

Half-Shot:
Fixes #205

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -40,6 +40,7 @@ import { Response } from "express";
 import { SlackRoomStore } from "./SlackRoomStore";
 import * as QuickLRU from "quick-lru";
 import PQueue from "p-queue";
+import { UserAdminRoom } from "./rooms/UserAdminRoom";
 
 const log = Logging.get("Main");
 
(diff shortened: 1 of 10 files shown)


Cadair:
Can you make this diff just the one diff? :angel: It largely looks fine, could probably do with at least some tests.

src/Main.ts
```diff
@@ -526,6 +533,23 @@ export class Main {
 
         const room = this.rooms.getByMatrixRoomId(ev.room_id);
         if (!room) {
+            const adminRoomUser = await this.datastore.getUserForAdminRoom(ev.room_id);
```
benparsons: took me a while to understand this whole section, a useful refactor could be to subdivide `onMatrixEvent` into several smaller methods and call them appropriately, rather than falling through.

auscompgeek:
I think this was accidentally opened against master instead of develop

PR #331, opened 2019-12-03
Add support for syncing channel creation and removal dynamically

Half-Shot:
Replaces #323 

Rewrote PR with changes from `nv/fm`.

This adds support to sync:
- Public channels
- Users

A future PR will allow us to sync private channels.


src/Main.ts
```diff
@@ -321,10 +327,19 @@ export class Main {
         return this.getGhostForSlack(message.user_id, teamDomain, teamId);
     }
 
-    public async getGhostForSlack(slackUserId: string, teamDomain: string, teamId?: string): Promise<SlackGhost> {
+    public async getGhostForSlack(slackUserId: string, teamDomain?: string, teamId?: string): Promise<SlackGhost> {
+        let domain: string;
```
benparsons: it feels weird to need to use another variable here, but I couldn't find a way without assertion...

```
if (!teamDomain && !teamId) {
            throw Error("Must provide either a teamDomain or a teamId");
        }
        if (! teamDomain) {
            teamDomain = await this.getTeamDomainForMessage({team_id: teamId});
        }
        const userId = this.getUserId(
            slackUserId.toUpperCase(),
            <string>teamDomain,
        );
```

feels nicer and needs no extra var, but needs an assertion - any way around this?

src/SlackEventHandler.ts
```diff
@@ -325,4 +341,24 @@ export class SlackEventHandler extends BaseSlackHandler {
...
+            throw Error("ignored");
+        }
+        if (event.type === "channel_created") {
+            // Note: Slack violates the usual stringness of 'channel' here.
+            const eventDetails = event as unknown as ISlackChannelAdded;
+            await this.main.teamSyncer.onChannelAdded(teamId, eventDetails.channel.id, eventDetails.channel.name, eventDetails.channel.creator);
+        } else if (event.type === "channel_deleted") {
+            await this.main.teamSyncer.onChannelDeleted(teamId, event.channel);
+        } else if (event.type === "team_join") {
+            const user = event.user as unknown as ISlackUser;
```
benparsons: What type is `event.type` that we need to use double assertion?

src/datastore/NedbDatastore.ts
```diff
@@ -214,6 +214,10 @@ export class NedbDatastore implements Datastore {
         return [];
     }
 
+    public async getPuppetMatrixUserBySlackId(teamId: string, slackId: string): Promise<null> {
+        return null;
```
benparsons: if we don't support nedb any longer do we need to keep stubbing these methods?

Half-Shot: Probably not. I may just chuck it out in another PR.

Half-Shot: Ah, but we use it for testing still...

src/SlackEventHandler.ts
Half-Shot: Good catch.

src/Main.ts
Half-Shot: Shortened it a bit, but we need the extra var because `teamDomain` is `string|undefined`

benparsons:
ok lgtm

PR #332, opened 2019-12-03
Membership Sync Slack <-> Matrix

Half-Shot:
This does several fancy things:

- Handles all the various ways the bridge needs to handle slack joins (it's surprisingly complex!).
- Handles leaves
- Handles invites
- Handles joining a matrix user's puppet upon joining a channel.

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -361,7 +361,7 @@ export class BridgedRoom {
 
     public async onMatrixMessage(message: any) {
         const puppetedClient = await this.main.clientFactory.getClientForUser(this.SlackTeamId!, message.user_id);
-        if (!this.slackWebhookUri && !this.botClient) { return false; }
+        if (!this.slackWebhookUri && !this.botClient && !puppetedClient) { return false; }
         const slackClient = puppetedClient || this.botClient;
         const user = this.main.getOrCreateMatrixUser(message.user_id);
         message = await this.stripMatrixReplyFallback(message);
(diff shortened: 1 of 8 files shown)


Cadair:
what about #293 ?! :stuck_out_tongue:

src/TeamSyncer.ts
```diff
@@ -273,15 +280,24 @@ export class TeamSyncer {
...
                 slack_channel_id: channelItem.id,
                 team_id: teamId,
             });
         } catch (ex) {
             log.error("Failed to provision new room dynamically:", ex);
+            return;
+        }
+
+        try {
+            await this.syncMembershipForRoom(roomId, channelItem.id, teamId, client);
```
benparsons: In general it doesn't seem correct to have have two calls to `syncMembershipForRoom`.

the logic is:

```
if (! existingroom) {
  // creation steps
}
syncMembershipForRoom()
```

But it seems like `creation steps` is so large it's better the way you have it. maybe 

```
if (! existingroom) {
  creationSteps()
}
syncMembershipForRoom()

function creationSteps() {
  // creation steps
}
```

could work, but not worth refactoring rn.

src/BridgedRoom.ts
```diff
@@ -457,6 +457,91 @@ export class BridgedRoom {
         return true;
     }
 
+    public async onSlackUserLeft(slackId: string) {
+        const ghost = await this.main.getGhostForSlack(slackId, undefined, this.slackTeamId);
+        await ghost.intent.leave(this.matrixRoomId);
+    }
+
+    public async onSlackUserJoin(slackId: string, wasInvitedBy?: string) {
```
benparsons: I could not find where this or `onSlackUserLeft` is called?

Half-Shot: Whoops, c&p fail.

Half-Shot:
yes :cry:

PR #333, opened 2019-12-04
Implement MSC2346

Half-Shot:
This is a proof of concept for [MSC2346](https://github.com/matrix-org/matrix-doc/2346) defines a schema for clients to identify which rooms are bridge to which network. This currently implements all the keys, but corners have been cut to make it work.


Half-Shot:
Hey there people on the tail end of 2025, closing this PR as development on the Slack bridge has been wound down now that the foundation is no longer running an instance. I'll work with the foundation on how we might want to communicate the status of this repository, but for now I'll close out any PRs that will not be landing.

PR #340, opened 2019-12-11
Support creating private chats on demand

Half-Shot:
This will create a private room if a user talks on the slack side but the room doesn't exist yet.

### DIFF
--- a/src/SlackRTMHandler.ts
+++ b/src/SlackRTMHandler.ts
@@ -208,14 +208,14 @@ export class SlackRTMHandler extends SlackEventHandler {
             await this.main.teamSyncer.onDiscoveredPrivateChannel(puppet.teamId, slackClient, chanInfo);
             return this.handleMessageEvent(event, puppet.teamId);
         }
-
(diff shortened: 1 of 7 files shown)


src/SlackRTMHandler.ts
```diff
@@ -228,9 +242,21 @@ export class SlackRTMHandler extends SlackEventHandler {
...
-            log.warn(`No room found for ${event.channel} and not sure how to create one`);
+            await Promise.all(otherGhosts.map((g) => g.intent.join(room_id)));
+            return this.handleMessageEvent(event, puppet.teamId);
+        }
+        log.warn(`No room found for ${event.channel} and not sure how to create one`);
+        return;
+    }
+
+    private async determineRoomName(chan: ConversationsInfo, otherGhosts: SlackGhost[],
+                                    puppet: PuppetEntry, client: WebClient): Promise<string|undefined> {
```
benparsons: why are we passing a PuppetEntry here?

benparsons: is there a way to have the linter report this?

Issue #354, opened 2020-02-19
TeamClient can get stuck with a broken token, which doesn't match what's in the DB


Half-Shot:
A user recently set up their team and tried to fetch the channels, but the requests kept failing with a `invalid_auth` error code. Restarting the bridge and allowing the team to be cached again fixed it. I wonder if we should try to refresh the team client upon these failures, just to be safe.

PR #355, opened 2020-02-19
Ensure team token is still valid

Half-Shot:
We need to regularly check that the access token cached is still valid, as a validation step. This should hopefully fix things like #354. We've also cheekily cranked up the number of channels returned from a `channels` call on the provisioner.

### DIFF
--- a/src/SlackClientFactory.ts
+++ b/src/SlackClientFactory.ts
@@ -64,16 +64,25 @@ export class SlackClientFactory {
         if (this.teamClients.has(teamId)) {
             const set = this.teamClients.get(teamId);
             // Check the auth on the client every AUTH_INTERVAL_MS, and if it fails, refetch the team.
-            if (set && Date.now() >= AUTH_INTERVAL_MS) {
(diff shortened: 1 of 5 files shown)


src/SlackClientFactory.ts
```diff
@@ -52,10 +62,33 @@ export class SlackClientFactory {
      */
     public async getTeamClient(teamId: string): Promise<WebClient> {
         if (this.teamClients.has(teamId)) {
-            return this.teamClients.get(teamId)!;
+            const set = this.teamClients.get(teamId);
+            // Check the auth on the client every AUTH_INTERVAL_MS, and if it fails, refetch the team.
+            if (set && Date.now() - set.lastTestTs < AUTH_INTERVAL_MS) {
+                // set has not expired
+                return set.client;
+            } else if (set) {
```
benparsons: Is there a condition where `this.teamClients.get(teamId)` could return null? Why are we checking `set` here?

Half-Shot: Yep, if the team hasn't been created yet, it will be null.

PR #364, opened 2020-03-18
Clarify path of config.yml in the docs

jaller94:
I got confused by the path of step 3 including the `config/` folder, but `config.yaml` not including the folder name.
When they both include the folder name, it's easier to tell where the new file should be placed.


jaller94:
Actually, `config.yaml` (not inside the `config` folder) was right.

PR #366, opened 2020-03-18
Minor wording suggestions in the docs

jaller94:
EX -> Example
The full word seems more accessible. I had not heard EX before.

ports -> port
As far as I can tell only one port is needed.

### DIFF
--- a/changelog.d/366.doc
+++ b/changelog.d/366.doc
@@ -1,0 +1,1 @@
+Minor wording changes in the docs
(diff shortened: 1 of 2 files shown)


changelog.d/366.doc
```diff
@@ -0,0 +1 @@
+Minor wording changes in the docs
```
Half-Shot:
```suggestion
Minor wording changes in `getting_started.md`
```

PR #368, opened 2020-03-18
Include server name in the Matrix users regex

### LINKED ISSUE #360: The registration file we generate doesn't have the server name in it
The users regex we generate should have the server part in it and it doesn't.

jaller94:
Fixes #360 by including the server name in the regex that determines which users the bridge controls.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -30,7 +30,7 @@ const cli = new Cli({
         reg.setHomeserverToken(AppServiceRegistration.generateToken());
         reg.setAppServiceToken(AppServiceRegistration.generateToken());
         reg.setSenderLocalpart("slackbot");
-        reg.addRegexPattern("users", `@${config.username_prefix}.*`, true);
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -30,7 +30,7 @@ const cli = new Cli({
         reg.setHomeserverToken(AppServiceRegistration.generateToken());
         reg.setAppServiceToken(AppServiceRegistration.generateToken());
         reg.setSenderLocalpart("slackbot");
-        reg.addRegexPattern("users", `@${config.username_prefix}.*`, true);
+        reg.addRegexPattern("users", `@${config.username_prefix}.*:${config.homeserver.server_name}`, true);
```
auscompgeek: Hm, shouldn't the server name be regex-escaped too?

Half-Shot: Oh maybe. Hopefully people don't have bizarre looking servernames, and if they do, they will know to fix it.

PR #372, opened 2020-03-19
Add documentation for Team Sync

### LINKED ISSUE #371: Add documentation for Team Sync feature
We want to add a page explaining what Team Sync does and how to configure it.

jaller94:
Fixes #371 
In the documentation I would like to call the feature Workspace Sync, because Teams seems like a deprecated Slack term to me.


jaller94:
There are two `TODO`s in the text which I didn't know to answer but felt like they were important.

While I had `alias_prefix` set to `slack_`, the resulting Matrix rooms did not have any aliases. Is this feature working?

docs/team_sync.md
```diff
@@ -0,0 +1,57 @@
+Workspace Sync
+==============
+
+The term Team is synonymously used for a Slack Workspace.
+
+The configuration option `team_sync` allows to replicate channels and users from Slack to Matrix before a user writes a message.
+
+To have a consistent experience across both sides of the bridge, it is recommended to have this feature enabled.
```
Half-Shot: I'd keep this sentence on the same line as previous.

docs/team_sync.md
```diff
@@ -0,0 +1,57 @@
...
+
+TODO: Why would you not have this enabled?
```
Half-Shot: Not sure, but let's leave this comment out of the markdown for now. We can open an issue for this if needed.

docs/team_sync.md
```diff
@@ -0,0 +1,57 @@
...
+      enabled: true
+  # Defaults for all other teams
+  all:
+    channels:
+      enabled: false
+```
+
+If not all channels shall be replicated, one may define an optional blacklist or whitelist of channel ids. If both a blacklist and a whitelist are defined, the blacklist takes precedence.
+
+The display name of the Matrix room will be identical to the one in Slack. Use the optional key `alias_prefix` to prevent conflicts with existing Matrix room aliases. Given the example above, a channel `#general` would get the alias address TODO.
```
Half-Shot: TODO?

docs/team_sync.md
```diff
@@ -0,0 +1,57 @@
...
+
+TODO: Why would you not have this enabled?
+
+## Table of supported events
+
+| Event | Without Team Sync | With Team Sync |
+|---|---|---|
+| A channel gets created | not replicated | the channel admin gets a message on how to bridge the channel |
+| A channel gets deleted | not replicated | immediately replicated |
+| User updates profile (e.g. Display name) | updated on their next message | immediately replicated |
```
Half-Shot: We also handle deleted users or new users joining the team, too.

docs/team_sync.md
jaller94: While I had set `alias_prefix="slack_"`, the resulting Matrix rooms did not have any aliases. Is this feature working?

jaller94: From reading the code, I assume it should be `#slack_general:your-homeserver.org`.

https://github.com/matrix-org/matrix-appservice-slack/blob/a2488ffaa5572d131d1f4d5dc7cd98e55cfe0a28/src/TeamSyncer.ts#L394

Half-Shot: Yep, and it should work :thinking:

jaller94: will do :)

docs/team_sync.md
jaller94: removed in 51186b6

Cadair:
This doesn't appear to have shown up in the built docs: https://matrix-appservice-slack.readthedocs.io/en/latest/

jaller94:
Do I have to link it somewhere?
Is the documentation based on `develop` or `master`? This is not in `master` yet.

Cadair:
"latest" should be built of develop and "stable" off master if we have done it right.

Cadair:
Ah the builds are failing because of a version conflict: https://readthedocs.org/projects/matrix-appservice-slack/builds/10657197/

PR #374, opened 2020-03-20
Correct ISlackEvent.user type

### LINKED ISSUE #373: Incorrect type annotation for user on ISlackEvent
Depending on the event type `user` is:
 * undefined,
 * an id string or
 * an ISlackUser (e.g. for the type `user_change`).

The current type annotation is incorrect.
https://github.com/matrix-org/matrix-appservice-slack/blob/1b5ce5ed70e9721e3d80b51247ee4492b50b1e7c/src/BaseSlackHandler.ts#L52

It's even misleading as in some cases the variable later gets re-casted to an incompatible type to correct for the false type.
https://github.com/matrix-org/matrix-appservice-slack/blob/1b5ce5ed70e9721e3d80b51247ee4492b50b1e7c/src/SlackEventHandler.ts#L361

jaller94:
Fixes #373

Depending on the event type `user` is:
 * undefined,
 * an id string or
 * an ISlackUser (e.g. for the type `user_change`).

### DIFF
--- a/src/SlackEventHandler.ts
+++ b/src/SlackEventHandler.ts
@@ -20,14 +20,15 @@ import { Main } from "./Main";
 import { Logging } from "matrix-appservice-bridge";
 const log = Logging.get("SlackEventHandler");
 
-interface ISlackEventChannelRenamed extends ISlackEvent {
+interface ISlackEventChannelRename extends ISlackEvent {
     // https://api.slack.com/events/channel_rename
     id: string;
     name: string;
     created: number;
 }
 
-interface ISlackEventTeamDomainChanged extends ISlackEvent {
+interface ISlackEventTeamDomainChange extends ISlackEvent {
+    // https://api.slack.com/events/team_domain_change
     url: string;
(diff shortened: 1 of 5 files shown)


src/SlackEventHandler.ts
```diff
@@ -20,14 +20,15 @@ import { Main } from "./Main";
 import { Logging } from "matrix-appservice-bridge";
 const log = Logging.get("SlackEventHandler");
 
-interface ISlackEventChannelRenamed extends ISlackEvent {
+interface ISlackEventChannelRename extends ISlackEvent {
     // https://api.slack.com/events/channel_rename
```
Half-Shot: We tend to put these in `/**` comments above the interface so that editors pick up the text. They don't tend to pick up the text below the interface declaration.

jaller94: I was already thinking that this would be cool as a documentation comment.
There were two examples like this in the file, that's why I continued the style I had found.

Will change this now.

PR #375, opened 2020-03-20
Remove ts-lint-disable where no-any is obsolete or does matter

jaller94:
There are some obsolete `ts-lint-disabled no-any` lines in the code and some where we'll probably want to replace any with something different in the long-term.

`no-any` is currently set to warning.

I've not touched any `event: any` lines as those are basically unpredictable responses from an API.

### DIFF
--- a/changelog.d/375.misc
+++ b/changelog.d/375.misc
@@ -1,0 +1,1 @@
+Enable code linting for no-any where it does matter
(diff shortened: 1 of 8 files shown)


Half-Shot:
>I've not touched any event: any lines as those are basically unpredictable responses from an API.

Yeah, at some point I admitted defeat on trying to make these typed but we should revisit it some day.

PR #377, opened 2020-03-20
Add user_change to the list of supported events

jaller94:
We are not actually listening to `user_change` events, so when a Slack user updates their display name, it does not get replicated to Matrix.

https://github.com/matrix-org/matrix-appservice-slack/blob/a2488ffaa5572d131d1f4d5dc7cd98e55cfe0a28/src/SlackRTMHandler.ts#L148-L150

This PR add `user_change` to the list of supported events which determines what Slack events we listen to.

name change-in-matrix


changelog.d/377.bug
```diff
@@ -0,0 +1 @@
+Fix Slack user updates (e.g. to their Display name) not getting immediately synced with Workspace Sync enabled
```
Half-Shot: `377.bugfix` :)

Issue #378, opened 2020-03-23
Bridge should report the number of users "using" the bridge via metrics.


Half-Shot:
For Matrix users, this should be defined by the following statements:
- They have sent a message/joined a room/caused some action to occur on the bridge.
- They have been active for at least N days in a given month. For example, talking on the 1st and then again on the 28th would consider them active as the timestamp between their first interation and their last interaction is significant. However, someone who only talked once is not to be considered active.
- OR they have an account that is puppeted by the bridge.

For slack users, this should be:
- They have sent a message/joined a channel/caused some action to occur on the bridge.
- They have been active for at least N days in a given month. For example, talking on the 1st and then again on the 28th would consider them active as the timestamp between their first interation and their last interaction is significant. However, someone who only talked once is not to be considered active.
- We should ignore slack bot users for now.

And just for fun, we should include the number of bridged rooms, again by activity:
- How many rooms have seen action, within a given period.

The metrics should include:
- How many active matrix users NOT being puppeted
- How many active matrix users being puppeted.
- How many rooms are active, by type of room.

jaller94:
**Work in progress**

New config variable:
```
# Amount of days that a user or a room needs to have bridged events to count as active
metrics_active_days_threshold: 2
```

**All gauges also have a label "team_id"**
```
# HELP bridge_active_users Amount of active users
# TYPE bridge_active_users gauge
bridge_active_users{remote=false} 34
bridge_active_users{remote=true} 27

# HELP bridge_user_puppets Amount of puppeted users on the remote side of the bridge
# TYPE bridge_user_puppets gauge
bridge_user_puppets 12

# HELP bridge_rooms Amount of active, bridged rooms per team (types are "channel" and "user")
# TYPE bridge_rooms gauge
bridge_active rooms{type="channel"} 20
bridge_active_rooms{remote="user"} 51
```

jaller94:
Solved in #380.

PR #380, opened 2020-03-23
WIP: Add metrics for active users and rooms

jaller94:
Possible solutions for #378


jaller94:
* I suggest to have separate tables for the metrics to avoid conflicts with the development of features.
* I suggest to have a users table to store if a user is remote and if they are a puppet.
* I assume that both a local and a remote user can be puppets and non-puppets. That's why I chose two booleans.
* I put the entire migration into one SQL query so it would succeed or fail entirely (The v1 migration did not do this).

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,26 @@
+import { IDatabase } from "pg-promise";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create schema
+    await db.none(`
+    CREATE TABLE metrics_users (
+        user_id TEXT NOT NULL,
+        remote BOOLEAN,
+        puppeted BOOLEAN
```
Half-Shot: again, might be best to link against the users table and infer if the user is puppeted perhaps?

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,26 @@
+import { IDatabase } from "pg-promise";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create schema
+    await db.none(`
+    CREATE TABLE metrics_users (
+        user_id TEXT NOT NULL,
+        remote BOOLEAN,
```
Half-Shot: and also this, probably can infer remote data from the users table also

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,26 @@
+import { IDatabase } from "pg-promise";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create schema
+    await db.none(`
+    CREATE TABLE metrics_users (
```
Half-Shot: In general, I'd expect this to be a table of (user_id, last_active_ts)

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,26 @@
...
+        user_id TEXT NOT NULL,
+        remote BOOLEAN,
+        puppeted BOOLEAN
+    );
+
+    CREATE TYPE room_type AS ENUM ('user', 'channel');
+    CREATE TABLE metrics_rooms (
```
Half-Shot: In general, I'd expect this to be a table of (room_id, last_active_ts)

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,26 @@
...
+        room_id TEXT NOT NULL,
+        type room_type
+    );
+
+    CREATE TABLE metrics_user_room_activities (
```
Half-Shot: Interesting to have a table of user+room activity, but I am wondering if it's strictly needed, and also if it's information we need to know.

src/Main.ts
```diff
@@ -277,6 +280,16 @@ export class Main {
             labels: ["outcome"],
             name: "remote_request_seconds",
         });
+        this.metrics.addGauge({
+            help: "count of active users",
+            labels: ["remote", "puppeted"],
```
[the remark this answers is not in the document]
jaller94: For the "remote" label, I want this to be true or false. This would give us both gauges.
```
bridge_active_users{remote=false} 34
bridge_active_users{remote=true} 27
```

Good call for pointing out we're interested in the number of puppets users, regardless of them being active or not. My next gauges would then be:
```
bridge_user_puppets{remote=false} 0
bridge_user_puppets{remote=true} 12
```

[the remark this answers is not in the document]
Half-Shot: Yep, makes sense to me. One last thing is that Slack users cannot puppet, so you can omit the remote field for the puppets metric :)

src/datastore/postgres/schema/v5.ts
jaller94: * I need `type` as active rooms are supposed to be split by their type.
* `last_active_ts` wouldn't allow me to answer "active for at least N days in a given month". My plan is to JOIN this table with `metrics_user_room_activities` which has timestamps with the granularity of a day.

src/datastore/postgres/schema/v5.ts
jaller94: * `last_active_ts` wouldn't allow me to answer "active for at least N days in a given month". My plan is to JOIN this table with `metrics_user_room_activities` which has timestamps with the granularity of a day.

src/datastore/postgres/schema/v5.ts
jaller94: One of my reasons for keeping the metrics separate is, if we had a constraint to functional data, deleting a user or room would reduce the team's usage of a system.

src/MatrixUser.ts
```diff
@@ -33,49 +43,35 @@ export class MatrixUser {
      * @param roomId The roomId to calculate the user's displayname for.
      */
     public getDisplaynameForRoom(roomId: string) {
```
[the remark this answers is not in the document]
jaller94: I added some unit tests for this method. Please have a look, if they are satisfactory.

src/MatrixUser.ts
```diff
@@ -33,49 +43,35 @@ export class MatrixUser {
...
-            // Disambiguate
-            return `${displayname} (${this.userId})`;
-        }
-
-        return displayname;
+        // Disambiguate, if the display name is used more than once.
+        return (matches.length > 1) ? `${displayname} (${this.userId})` : displayname;
     }
 
     public getAvatarUrlForRoom(roomId: string) {
```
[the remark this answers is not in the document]
jaller94: I added some unit tests for this method. Please have a look, if they are satisfactory.

src/datastore/postgres/schema/v5.ts
jaller94: This table came out of the requirement to tell "active [rooms/users] for at least N days in a given month".
Each combination of (user, bridged room and date) will be a unique row, indicating one or more activities.

src/Main.ts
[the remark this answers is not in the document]
jaller94: I added `team_id` as a label to all new metrics.

src/Main.ts
```diff
@@ -277,6 +280,16 @@ export class Main {
...
+            name: METRIC_ACTIVE_USERS,
+        });
+        this.metrics.addGauge({
+            help: "count of active bridged rooms",
+            labels: ["type"],
```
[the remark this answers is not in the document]
jaller94: * I added the two currently known values to the HELP text.
* `team_id` was added as a label.

src/MatrixUser.ts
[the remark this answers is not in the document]
Half-Shot: Did you commit them?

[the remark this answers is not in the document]
jaller94: weren't pushed until a second ago 🤦‍♂👍

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,28 @@
...
+        );
+
+        CREATE TABLE metrics_user_room_activities (
+            user_id TEXT NOT NULL,
+            room_id TEXT NOT NULL,
+            date DATE,
+            CONSTRAINT cons_user_room_activities_unique UNIQUE(user_id, room_id, date),
+            CONSTRAINT cons_user_room_activities_room
+                FOREIGN KEY(room_id) REFERENCES metrics_rooms(room_id)
+        );
```
Half-Shot: Hmm, so I guess the reason for having a `DATE` object is the reduced time accuracy costs us less in storage, as a user+room_id can only store one item per day. I guess this is fine, although it might be nice to have greater accuracy for debugging purposes. Doing that would require more cleanup though, so in theory this is fine.

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,28 @@
+import { IDatabase } from "pg-promise";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create schema
+    await db.none(`
+        CREATE TABLE metrics_users (
```
Half-Shot: Looking at this again, this feels useless if we already have a users table. I'm minded to drop it.

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,28 @@
...
+            user_id TEXT NOT NULL PRIMARY KEY,
+            remote BOOLEAN NOT NULL
+        );
+
+        CREATE TYPE room_type AS ENUM ('user', 'channel');
+        CREATE TABLE metrics_rooms (
```
Half-Shot: Same for this, we might just need the metrics table.

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+            "SELECT team_id, user_id, COUNT(*) AS active_days" +
+            "FROM metrics_user_room_activities" +
+            "WHERE date >= DATE_SUB(CURRENT_DATE, INTERVAL ${historyLengthInDays} DAYS)" +
+            "INNER JOIN metrics_users ON metrics_user_room_activities.user_id = metrics_users.user_id" +
+            "GROUP BY team_id, user_id" +
+            "HAVING active_days > ${activityThreshholdInDays}",
+            { activityThreshholdInDays, historyLengthInDays },
+        )).map((u) => ({
+            teamId: u.matrixuser,
+            activeRooms: u.activeRooms,
```
Half-Shot: activeUsers, surely.

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+        //     ]
+        // };
+    }
+
+    public async getActiveUsersPerTeam(activityThreshholdInDays = 2, historyLengthInDays = 30): Promise<any> {
+        return (await this.postgresDb.manyOrNone(
+            "SELECT team_id, user_id, COUNT(*) AS active_days" +
+            "FROM metrics_user_room_activities" +
+            "WHERE date >= DATE_SUB(CURRENT_DATE, INTERVAL ${historyLengthInDays} DAYS)" +
+            "INNER JOIN metrics_users ON metrics_user_room_activities.user_id = metrics_users.user_id" +
```
Half-Shot: I think there must be a cunning way to link to `users`

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+            "GROUP BY team_id, user_id" +
+            "HAVING active_days > ${activityThreshholdInDays}",
+            { activityThreshholdInDays, historyLengthInDays },
+        )).map((u) => ({
+            teamId: u.matrixuser,
```
Half-Shot: This isn't a team_id?

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+        // };
+    }
+
+    public async getActiveUsersPerTeam(activityThreshholdInDays = 2, historyLengthInDays = 30): Promise<any> {
+        return (await this.postgresDb.manyOrNone(
+            "SELECT team_id, user_id, COUNT(*) AS active_days" +
+            "FROM metrics_user_room_activities" +
+            "WHERE date >= DATE_SUB(CURRENT_DATE, INTERVAL ${historyLengthInDays} DAYS)" +
+            "INNER JOIN metrics_users ON metrics_user_room_activities.user_id = metrics_users.user_id" +
+            "GROUP BY team_id, user_id" +
```
Half-Shot: Hmm, we just want the count of users mapped against the team_id, rather than the user_ids themselves. Although, we can just count it so it's probably not important

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+            "HAVING active_days > ${activityThreshholdInDays}",
+            { activityThreshholdInDays, historyLengthInDays },
+        )).map((u) => ({
+            teamId: u.matrixuser,
+            activeRooms: u.activeRooms,
+        }));
+        // return {
```
Half-Shot: Actually, this response format doesn't match what the statement above return. The response format below is actually what we wanted though :)

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,69 @@ export class PgDatastore implements Datastore {
...
+        //     "ABCDEFGH": [
+        //         "#general:localhost|ABCDEFG",
+        //         "#random:localhost|BCDEFGH",
+        //         "#finances:localhost|CDEFGHI",
+        //     ]
+        // };
+    }
+
+    public async getActiveUsersPerTeam(activityThreshholdInDays = 2, historyLengthInDays = 30): Promise<any> {
+        return (await this.postgresDb.manyOrNone(
```
Half-Shot: Most of the comments below apply to the above statement too.

src/datastore/Models.ts
```diff
@@ -100,7 +107,15 @@ export interface Datastore {
     getPuppetsByMatrixId(userId: string): Promise<PuppetEntry[]>;
     getPuppetedUsers(): Promise<PuppetEntry[]>;
 
+    // Admin rooms
     getUserAdminRoom(matrixId: string): Promise<string|null>;
     getUserForAdminRoom(roomId: string): Promise<string|null>;
     setUserAdminRoom(matrixuser: string, roomId: string): Promise<void>;
+
+    // Metrics
+    getActiveRoomsPerTeam(activityThreshholdInDays?: number, historyLengthInDays?: number): Promise<any>;
```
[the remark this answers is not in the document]
jaller94: That means the names aren't great in the first place. :/
Will add the doc strings but feel free to suggest better names as I will look for other names.

src/datastore/postgres/schema/v5.ts
jaller94: I removed the metrics_users table.

src/datastore/postgres/schema/v5.ts
jaller94: I removed the metrics_users table.

src/tests/unit/MatrixUserTest.ts
```diff
@@ -20,20 +20,87 @@ import { MatrixUser } from "../../MatrixUser";
...
+    describe("getDisplayName", () => {
+        it("returns the user_id when no displayName is given", () => {
+            const getStoredEvent = getStoredEventGenerator([]);
+            const user = new MatrixUser({ getStoredEvent } as Main, { user_id: "@alice:localhost" });
+            expect(user.getDisplaynameForRoom("")).to.equal("@alice:localhost");
+        });
+        it("returns the displayName if one is given", () => {
+            const getStoredEvent = getStoredEventGenerator([
+                { id: "@alice:localhost", displayName: "Alice" },
+                { displayName: "Hatmaker" },
```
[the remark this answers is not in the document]
jaller94: The id is only strictly needed when the user is important. In this case the important feature is the displayName which differs.
We could enforce `id` to be a string to better represent join events, but at the moment there is no functional reason to do so.

src/datastore/postgres/schema/v1.ts
```diff
@@ -3,38 +3,40 @@ import { IDatabase } from "pg-promise";
 // tslint:disable-next-line: no-any
```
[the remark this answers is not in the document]
jaller94: Reverted `v1.ts`
It wasn't just whitespaces, but it's unrelated to this PR.

src/datastore/Models.ts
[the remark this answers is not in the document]
jaller94: Replaced with an new type called `ActiveRoomEntry`

jaller94:
Matrix users don't seem to be included in the `users` table.
We need to associate them with a teamId for the metrics.

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,16 @@
...
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create schema
+    await db.none(`
+        CREATE TABLE metrics_activities (
+            user_id TEXT NOT NULL,
+            room_id TEXT NOT NULL,
+            date DATE,
+            CONSTRAINT cons_activities_unique UNIQUE(user_id, room_id, date),
+            CONSTRAINT cons_activities_user FOREIGN KEY(user_id) REFERENCES users(userid),
```
Half-Shot: Remove this.

src/datastore/postgres/schema/v5.ts
```diff
@@ -0,0 +1,16 @@
...
+            CONSTRAINT cons_activities_room FOREIGN KEY(room_id) REFERENCES rooms(id)
```
Half-Shot: Remove this.

jaller94: removed the foreign keys

src/Main.ts
```diff
@@ -277,6 +285,27 @@ export class Main {
...
+            name: METRIC_ACTIVE_ROOMS,
+        });
+        this.metricPuppets = this.metrics.addGauge({
+            help: "Amount of puppeted users on the remote side of the bridge",
+            labels: ["team_id"],
+            name: METRIC_PUPPETS,
+        }) as Gauge;
+
+        const ONE_HOUR = 60 * 60 * 1000;
+        setInterval(() => {
```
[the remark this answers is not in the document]
jaller94: The interval now gets stored and cleaned up.
23121a0fda7edeb441a1957ad794e1fec103ada2

src/datastore/postgres/schema/v5.ts
jaller94: removed the foreign key

src/datastore/postgres/PgDatastore.ts
```diff
@@ -292,6 +292,50 @@ export class PgDatastore implements Datastore {
...
+    public async getActiveRoomsPerTeam(activityThreshholdInDays = 2, historyLengthInDays = 30): Promise<ActiveRoomEntry[]> {
+        return (await this.postgresDb.manyOrNone(
+            "SELECT room_id, rooms.json::json->>'slack_team_id' AS team_id, rooms.json::json->>'slack_type' AS slack_type, COUNT(DISTINCT date) AS active_days " +
+            "FROM metrics_activities " +
+            "INNER JOIN rooms ON metrics_activities.room_id = rooms.id " +
+            "WHERE date_part('days', age(date)) < ${historyLengthInDays} " +
+            "GROUP BY room_id, team_id, room_id, slack_type " +
+            "HAVING COUNT(DISTINCT date) >= ${activityThreshholdInDays};",
+            { activityThreshholdInDays, historyLengthInDays },
+        )).map((u) => ({
```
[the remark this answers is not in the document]
jaller94: Moved the transformation into the datastore function.
91468d349992268b7e30f0c9fedfa5e842dd7a29

src/SlackClientFactory.ts
```diff
@@ -220,6 +225,9 @@ export class SlackClientFactory {
             const auth = (await slackClient.auth.test()) as AuthTestResponse;
             const user = (await slackClient.users.info({user: auth.user_id})) as UsersInfoResponse;
             log.debug("Created new team client for", teamInfo.team.name);
+            if (this.updatePuppetCount) {
+                this.updatePuppetCount(teamInfo.team.id, 1);
+            }
```
Half-Shot: We're on Typescript 3.8, which means we can use [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)
```suggestion
            this.updatePuppetCount?.(teamInfo.team.id, 1);
```

PR #381, opened 2020-03-24
Upgrade various low-risk dependencies

jaller94:
These upgrades are unlikely to break functionality and some manual testing showed no difference.


jaller94:
I tested sending messages, creating new users on both sides and changing display names.
I did not test connecting a new team/workspace.

PR #382, opened 2020-03-24
Upgrade uuid dependency

jaller94:
[Deep requires have been deprecated for this package](https://www.npmjs.com/package/uuid#deep-requires-now-deprecated)


Half-Shot:
Wow, it's that old! Thanks for fixing

PR #384, opened 2020-03-26
Add Launch config for VS Code

jaller94:
Adds a Launch config for VS Code, allowing to quickly start and inspect the bridge inside of VS Code.

* Not sure, if we want to have this in the repo, as it is specific to one editor.
* I locally had picked port 9999. It's unfortunate that this is hard-coded in this PR, but I didn't want to make it flexible before I know there is an interest in this PR at all.
* SourceMaps need to be enabled for breakpoints to work.

Screenshot_2020-03-26_14-47-29

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,22 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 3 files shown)


Half-Shot:
> I locally had picked port 9999. It's unfortunate that this is hard-coded in this PR, but I didn't want to make it flexible before I know there is an interest in this PR at all.

Flexibility would be nice, but I'm not sure how you can do that.

Also yay for sourcemaps.

PR #385, opened 2020-03-27
Add error and debug logs to .gitignore

jaller94:
Is there a command line option to place them in the ignored `logs` folder?
I find it irritating that the default path for logs isn't listed in `.gitignore`.


Half-Shot:
Indeed :/. The default logging configuration is probably unhelpful.

PR #389, opened 2020-03-31
Improving the installation docs and config.sample.yaml 

jaller94:
* In the docs, [be explicit which config fields need to be changed](https://github.com/matrix-org/matrix-appservice-slack/pull/389/commits/f505114ec12b59e7042bef113e11d4a9a2ee7c2f).
* Simplify config.sample.yaml
  * Move `homeserver`, `db` and `matrix_admin_room` to the top. They will likely get changed by an admin.
  * Comment out `tls` and `oauth2`. They are optional, and `tls` prevents the server from starting in the former state.
* Rename the page "Workspace Sync" to the more common name "Team Sync"


README.md
```diff
@@ -33,7 +33,3 @@ has the time to spare. If you want to work on the bridge, please see
 Please come visit us in the 
 [support room](https://matrix.to/#/#matrix_appservice_slack:cadair.com/) if you
 have any questions.
-
-# Documentation
-
-This documentation is hosted on [Read the Docs](https://matrix-appservice-slack.rtfd.io/).
```
anoadragon453, line 39: I know we have a badge for it, but it doesn't seem like a bad idea to have a second link to the documentation.

changelog.d/389.doc
```diff
@@ -0,0 +1 @@
+Improve the installation docs and config.sample.yaml
```
anoadragon453, line 1:
```suggestion
Improve the installation docs and the sample config file
```

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
-# Prefix of incoming requests to strip when handing. This is NOT the bind host
-  #NOTE: Unlike most of the other urls, this one cannot use localhost,
-  #as this one must be publicly visible to the slack API
-inbound_uri_prefix: "https://my.server.here:9898/"
+homeserver: # Required
+  # The domain name of your homeserver
+  server_name: my.server.here
+  # The URL which the bridge server can reach the homeserver on.
+  # You can use localhost so long as the bridge and homeserver are
+  # hosted on the same machine
```
anoadragon453, line 6: This statement will probably help a lot of people, but *might* not be technically accurate for others.

However, those with changed environments probably know what they're doing, so I think this is fine.

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+  # 8008 is assumed to be the port the homeserver listens on
```
anoadragon453, line 7: I could see this potentially confusing somebody, putting `8008` at the end of `https://my.domain.com` or something.

I would probably just remove this line.

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+  url: http://localhost:8008
+  # The public facing url for media on your homeserver.
+  # This is usually the public url of your homeserver.
+  # Optional, will use `url` by default.
+  # media_url: "http://my.server.here"
+  # Optional. The maximum size of a uploaded file to Matrix in bytes. No limit by default
+  # max_upload_size: 104857600
```
anoadragon453, line 14: Could you either:

* Remove the space before commented out option names or
* Put a newline between options

as I find it hard to notice that there are multiple options in this comment blob.

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+  # The domain name of your homeserver
+  server_name: my.server.here
+  # The URL which the bridge server can reach the homeserver on.
+  # You can use localhost so long as the bridge and homeserver are
+  # hosted on the same machine
+  # 8008 is assumed to be the port the homeserver listens on
+  url: http://localhost:8008
+  # The public facing url for media on your homeserver.
+  # This is usually the public url of your homeserver.
+  # Optional, will use `url` by default.
```
anoadragon453, line 11:
```suggestion
  # Optional. The value of `url` will be used by default.
```

I initially read this as media_url will be set to `"url"`.

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+  # media_url: "http://my.server.here"
+  # Optional. The maximum size of a uploaded file to Matrix in bytes. No limit by default
+  # max_upload_size: 104857600
+  # Optional. Used to specify the port of the appservice in the config, rather than the cli.
```
anoadragon453, line 15:
```suggestion
  # Optional. Used to specify the port of the appservice in the config, rather than the command line.
```

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+  # If this is defined, it will **override** the port given in the process arguments.
+  # appservice_port: 9999
 
-# Both of the above options are optional if using RTM (see below), required otherwise.
+# Settings for database connection
+db:
+  # Which database engine to use
+  engine: "postgres" # "postgres" or "nedb". If using "nedb", do not define connectionString
```
anoadragon453, line 22: Will `connectionString` be unused or does the user have to comment out `connectionString` if using nedb?

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
-# TLS files to be used when handing slack requests
+# Use a matrix room to issue link and unlink commands to the bridge
 # Optional
-tls:
-  key_file: /path/to/tls.key
-  crt_file: /path/to/tls.crt
+matrix_admin_room: "!aBcDeF:matrix.org"
+
+# The prefix to give Slack users on the Matrix server
+username_prefix: "slack_" # Required
```
anoadragon453, line 31: You mention both `Required` and `Optional` in places, but I think you can get away with just one.

config/config.sample.yaml
```diff
@@ -1,17 +1,40 @@
...
+
+# TLS files to be used when handling Slack requests
+# Optional
```
anoadragon453, line 34:
```suggestion
# Optional. TLS files to be used when handling Slack requests.
```

To match other lines.

config/config.sample.yaml
```diff
@@ -22,35 +45,25 @@ rtm:
   enable: true
   logging: "silent" # Logging level specific to RTM traffic.
 
+## Start: Optional options if using RTM API, required otherwise.
```
anoadragon453, line 48: Nowhere in this document does it explain what "RTM" stands for.

config/config.sample.yaml
```diff
@@ -22,35 +45,25 @@ rtm:
...
+
+# Port for incoming slack requests from webhooks and event API messages
+slack_hook_port: 9898
+# Prefix of incoming requests to strip when handing. This is NOT the bind host
```
anoadragon453, line 52: handing?

config/config.sample.yaml
anoadragon453, line 48: This section boundary is a weird notation and not used anywhere else in this file. As it only applies to two options, can we just put the optional explanation on each?

README.md
jaller94: There's also one in "Setting up". This was the third link to the same docs.

config/config.sample.yaml
jaller94: It's a bit easier with syntax highlighting.
All my code editors put a space after the comment character, so it's very annoying to maintain comments without the space.
Adding newlines makes the file length significantly longer and less readable when introducing a newline consistently between all properties.

anoadragon453: These are just suggestions that we apply to [synapse's config](https://github.com/matrix-org/synapse/blob/develop/docs/sample_config.yaml), which while monumentally longer, we don't receive complaints about the length.

If you don't want to apply them though, that's fine - it's only a suggestion.

src/Main.ts
```diff
@@ -153,7 +153,7 @@ export class Main {
         }
 
         let bridgeStores = {};
-        const usingNeDB = config.db === undefined;
+        const usingNeDB = config.db === undefined || config.db?.engine === "nedb";
```
jaller94, line 156: @Half-Shot Explicitly setting `config.db.engine` to `"nedb"` never worked, did it?

Also, do you agree with this change to format the sample file like Synapses sample with a lot more spacing for comments? Anoa suggested it and I like it.

Half-Shot: >Also, do you agree with this change to format the sample file like Synapses sample with a lot more spacing for comments? Anoa suggested it and I like it.

Spacing never hurt anyone, I'm all for it.

>Explicitly setting config.db.engine to "nedb" never worked, did it?

It should do?

config/config.sample.yaml
```diff
@@ -1,155 +1,221 @@
...
+
+# Optional. Set the bot's profile on startup
+bot_profile:
+  # Optional. Set the displayname.
+  #
+  displayname: "Slack Bridge"
+  
+  # Optional. Set the avatar.
+  #
+  avatar_url: "mxc://half-shot.uk/ea64c71ee946ca2f61379abefe2c7d977d276fbb"
```
Half-Shot, line 221: We've since added new options to the config, can you double check we have all of them included in this file?

Half-Shot (approved):
I'm generally happy that this is an improvement, thank you for taking this on!

src/Main.ts
jaller94: The comments in the sample config and the schema list "nedb" as a valid value.

config/config.sample.yaml
Half-Shot: I don't see anything in there for the encryption stuff?

PR #397, opened 2020-04-23
Add configuration to limit teams and rooms counts

Half-Shot:
~~This is still missing functionality to handle the removal of team clients~~ Punting to https://github.com/matrix-org/matrix-appservice-slack/issues/399

This PR does 4 things:
- It limits the number of bridged rooms that can be created by config.
- It limits the number of teams that can be connected.
- It exposes a flag for whether the configured instance allows private rooms to be bridged to channels.
- It exposes these limits in a config endpoint.

### DIFF
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -102,3 +102,9 @@ team_sync:
 
 provisioning:
   enabled: true
+  # Should the bridge allow public rooms to be bridged
+  require_public_room: true
+  limits:
+    room_count: 20
+    team_count: 1
+
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -102,7 +102,7 @@ team_sync:
 
 provisioning:
   enabled: true
-  # Should the bridge allow public rooms to be bridged
+  # Should the bridge deny users bridging channels to private rooms.
   require_public_room: true
   limits:
     room_count: 20
--- a/src/IConfig.ts
+++ b/src/IConfig.ts
@@ -82,7 +82,7 @@ export interface IConfig {
 
     provisioning?: {
         enable: boolean;
-        require_public_room: boolean;
+        require_public_room?: boolean;
         limits?: {
             team_count?: number;
             room_count?: number;
--- a/src/IConfig.ts
+++ b/src/IConfig.ts
@@ -84,4 +84,9 @@ export interface IConfig {
         enable: boolean;
         auth_callbck: string;
     };
+
+    limits?: {
+        team_count?: number;
+        room_count?: number;
+    }
 }
--- a/src/IConfig.ts
+++ b/src/IConfig.ts
@@ -82,11 +82,10 @@ export interface IConfig {
 
     provisioning?: {
         enable: boolean;
-        auth_callbck: string;
+        require_public_room: boolean;
+        limits?: {
+            team_count?: number;
+            room_count?: number;
+        }
     };
-
-    limits?: {
-        team_count?: number;
-        room_count?: number;
-    }
 }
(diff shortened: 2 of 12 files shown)


config/config.sample.yaml
```diff
@@ -102,3 +102,9 @@ team_sync:
 
 provisioning:
   enabled: true
+  # Should the bridge allow public rooms to be bridged
+  require_public_room: true
```
jaller94, line 106: The comment says "should?", the variable name says "require".
This is confusing.

src/IConfig.ts
```diff
@@ -82,6 +82,10 @@ export interface IConfig {
 
     provisioning?: {
         enable: boolean;
-        auth_callbck: string;
+        require_public_room: boolean;
```
jaller94, line 85: `require_public_room` is not marked required in `config/slack-config-schema.yaml`

src/Main.ts
```diff
@@ -810,7 +811,20 @@ export class Main {
         }
         const port = this.config.homeserver.appservice_port || cliPort;
         this.bridge.run(port, this.config, this.appservice);
-        const roomListPromise = this.bridge.getBot().getJoinedRooms() as Promise<string[]>;
+        let joinedRooms: string[]|null = null;
+        while(joinedRooms === null) {
```
jaller94, line 815: I wish there was an event handler for this. While I guess the async function || timeout are achieving the same, this looks scary. (thinking never-ending loop or 100% CPU core usage).

The more I read through this, the less it concerns me though.. looks like it's safe.

src/SlackHookHandler.ts
```diff
@@ -341,6 +341,23 @@ export class SlackHookHandler extends BaseSlackHandler {
...
+                    log.warn(`User ${response.user_id} tried to add a new team ${response.team_id} but the team limit was reached`);
+                    try {
+                        const tempClient = await this.main.clientFactory.createTeamClient(response.access_token);
+                        await tempClient.slackClient.auth.revoke();
+                    } catch (ex) {
+                        log.warn(`Additionally failed to revoke the token:`, ex);
+                    }
+                    return {
+                        code: 403,
+                        // Not using templaes to avoid newline awfulness.
```
jaller94, line 355:
```suggestion
                        // Not using templates to avoid newline awfulness.
```

src/SlackHookHandler.ts
```diff
@@ -365,7 +382,7 @@ export class SlackHookHandler extends BaseSlackHandler {
                 // Not using templaes to avoid newline awfulness.
                 // tslint:disable-next-line: prefer-template
                 html: "<h2>Integration Failed</h2>\n" +
-                "<p>Unfortunately your channel integration did not go as expected...</p>",
+                `<p>Unfortunately your ${room ? "channel integration" : "account" } did not go as expected...</p>`,
```
jaller94, line 385:
```suggestion
                `<p>Unfortunately, your ${room ? "channel integration" : "account" } did not go as expected...</p>`,
```

src/SlackHookHandler.ts
```diff
@@ -365,7 +382,7 @@ export class SlackHookHandler extends BaseSlackHandler {
                 // Not using templaes to avoid newline awfulness.
```
jaller94, line 382:
```suggestion
                // Not using templates to avoid newline awfulness.
```

src/Main.ts
Half-Shot: I believe this is safe. The CPU won't spin while waiting for the await to complete, so there is no harm.

Half-Shot:
>Admin rooms are not in the rooms table, right?

Correct.

>We do remove rooms from the rooms table when they get unlinked, right?

Correct

PR #402, opened 2020-05-06
Support room upgrades

Half-Shot:
This change makes it so that the bridge follows m.room.tombstone events

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -170,6 +170,12 @@ export class Main {
                 },
                 onUserQuery: () => ({}), // auto-provision users with no additional data
             },
+            roomUpgradeOpts: {
+                consumeEvent: true,
+                migrateGhosts: true,
+                onRoomMigrated: this.onRoomUpgrade.bind(this),
+                migrateStoreEntries: false,
+            },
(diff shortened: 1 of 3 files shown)


src/Main.ts
```diff
@@ -1135,6 +1141,23 @@ export class Main {
...
+        log.info(`Room has been upgraded from ${oldRoomId} to ${newRoomId}`);
+        const bridgedroom = this.rooms.getByMatrixRoomId(oldRoomId);
+        const adminRoomUser = await this.datastore.getUserForAdminRoom(oldRoomId);
+        if (bridgedroom) {
+            log.info("Migrating channel");
+            this.rooms.removeRoom(bridgedroom);
+            bridgedroom.migrateToNewRoomId(newRoomId);
+            this.rooms.upsertRoom(bridgedroom);
+            await this.datastore.upsertRoom(bridgedroom);
+        } else if (adminRoomUser) {
```
jaller94, line 1154: If the admin room cannot be a bridged room, feel free to merge this.

Half-Shot: It really shouldn't be, no :p

PR #403, opened 2020-05-07
Allow bridging of private channels via the provisioner

Half-Shot:
This requires us to determine the identity of the caller in order to check if they are *in* the channel. Luckily, there is an endpoint for that :)

### DIFF
--- a/src/Provisioning.ts
+++ b/src/Provisioning.ts
@@ -87,6 +87,20 @@ export class Provisioner {
         return (currentCount >= this.main.config.provisioning?.limits?.room_count);
     }
 
+    private async determineSlackIdForRequest(matrixUserId, teamId) {
+        const matrixUser = await this.main.datastore.getMatrixUser(matrixUserId);
+        if (matrixUser === null) {
+            throw Error('No users found');
+        }
+        const accounts: {[userId: string]: {team_id: string}} = matrixUser.get("accounts");
+        for (const [key, value] of Object.entries(accounts)) {
+            if (value.team_id === teamId) {
+                return key;
+            }
+        }
+        return null;
+    }
+
     @command()
     private async getconfig(_, res) {
         const hasRoomLimit = this.main.config.provisioning?.limits?.room_count;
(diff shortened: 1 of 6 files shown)


src/Provisioning.ts
```diff
@@ -150,25 +164,25 @@ export class Provisioner {
...
         }
         const team = await this.main.datastore.getTeam(teamId);
         if (team === null) {
             throw Error("No team token for this team_id");
         }
         const cli = await this.main.clientFactory.getTeamClient(teamId);
         try {
-            const response = (await cli.conversations.list({
+            let types = "public_channel";
+            if (this.main.config.provisioning?.allow_private_channels !== false) {
```
jaller94, line 178: double negation

Wouldn't `allow_private_channels === true` or `allow_private_channels` be better and more natural to read?

src/Provisioning.ts
```diff
@@ -87,6 +87,20 @@ export class Provisioner {
         return (currentCount >= this.main.config.provisioning?.limits?.room_count);
     }
 
+    private async determineSlackIdForRequest(matrixUserId, teamId) {
+        const matrixUser = await this.main.datastore.getMatrixUser(matrixUserId);
+        if (matrixUser === null) {
+            throw Error('No users found');
```
jaller94, line 93: Nit-pick: User**s**? We're just looking for one user.

src/Provisioning.ts
```diff
@@ -87,6 +87,20 @@ export class Provisioner {
         return (currentCount >= this.main.config.provisioning?.limits?.room_count);
     }
 
+    private async determineSlackIdForRequest(matrixUserId, teamId) {
+        const matrixUser = await this.main.datastore.getMatrixUser(matrixUserId);
+        if (matrixUser === null) {
```
jaller94, line 92: I prefer `(!matrixUser)`.
btw. getMatrixUser's signatures don't match but TypeScript doesn't complain. Some implementations return undefined, others null.

Screenshot_2020-05-07_15-12-18

src/Provisioning.ts
Half-Shot: Ah, careful here. `allow_private_channels` is considered truthy unless explicitly false. All other values are true.

Half-Shot:
@jaller94 I didn't actually modify this PR yet :D

src/Provisioning.ts
Half-Shot: We should also be returning null.

src/Provisioning.ts
Half-Shot: Ah, because MatrixUser is any, so essentially types are disabled here.

Half-Shot:
>fakeDatastore still has the more general signature getMatrixUser(userId: string): Promise<any>.

Since it's not implemented yet, let's leave it for when that day comes.

PR #404, opened 2020-05-08
Fix error caused by a failed catch statement

Half-Shot:


Half-Shot:
(In the future, while the gifs are cute, it makes your response harder to read with a doggo head twisting away)

>Why are we catching so many errors, just to throw them with a different trace, making them less debuggable.

So, really what I should have done is logged the error but throw a fresh error. Throwing errors with a different message *is* useful because it gives useful context, for example:

Could not send message: Failed to get a slack team client

is more useful than

Could not send message: Slack request to /auth faiiled.

Or to put it another way, the caller should not need to know why a request failed but simply that the request did fail. We should however log the details to ensure it can be debugged by developers, so I will amend the PR.

PR #405, opened 2020-05-08
Pin Dockerfile node version

Half-Shot:
I have a suspicion that pg-promise broke on Node v14, as we're seeing people complaining about being unable to use postgres on the latest image, which is using Node v14.

To fix, I've updated pg-promise to a version compatible with Node v14, and also pinned us to Node v14 in order to ensure we don't get caught out next time. At some point, we should include Node v14 on our pipeline.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "nedb": "^1.8.0",
     "node-emoji": "^1.10.0",
     "p-queue": "^6.3.0",
-    "pg-promise": "^10.4.4",
+    "pg-promise": "^10.5.3",
     "quick-lru": "^5.0.0",
     "randomstring": "^1",
     "request-promise-native": "^1.0.8",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A Matrix <--> Slack bridge",
   "main": "app.js",
   "scripts": {
-    "postinstall": "npm run build",
+    "prepare": "npm run build",
```
jaller94, line 7: This will also run on `npm install`. In my opinion `prepack` would be a better script key to build the code.
https://docs.npmjs.com/misc/scripts

Half-Shot: I disagree. This will run on npm install, so when people install the package by cloning to their system. This saves an extra instruction of running `npm run build`

Half-Shot:
>As a developer, I don't expect the repo's build scripts to run when I run npm install.

It happens more than you think :wink:

PR #416, opened 2020-05-21
Fix team_sync indentation in sample config

mmilata:
I think `users` should be on the same nesting level as `channels`.


erdnaxeli:
Yeah I noticed this too and I think you're right.

PR #418, opened 2020-05-26
Fix logic error

### LINKED ISSUE #417: Logic error in commit 084786a
Line **53** in` src/rooms/UserAdminRoom.ts` should be:

 `if (!this.main.oauth2 || !this.main.config.puppeting?.enabled) {`

With the way it is currently, puppeting won't work if "enabled" is set in the config to "true".

Half-Shot:
Fixes #417

### DIFF
--- a/src/rooms/UserAdminRoom.ts
+++ b/src/rooms/UserAdminRoom.ts
@@ -50,7 +50,7 @@ export class UserAdminRoom {
     }
 
     public async handleLogin() {
-        if (!this.main.oauth2 || this.main.config.puppeting?.enabled) {
+        if (!this.main.oauth2 || !this.main.config.puppeting?.enabled) {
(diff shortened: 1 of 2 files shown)


src/rooms/UserAdminRoom.ts
```diff
@@ -50,7 +50,7 @@ export class UserAdminRoom {
     }
 
     public async handleLogin() {
-        if (!this.main.oauth2 || this.main.config.puppeting?.enabled) {
+        if (!this.main.oauth2 || !this.main.config.puppeting?.enabled) {
```
jaller94, line 53: This condition holds a lot of complexity with the negation of an optional object with a boolean inside.
But according to my testground, it's now doing what you expect it to do.
Screenshot_2020-05-26_17-22-20

PR #419, opened 2020-05-27
Refactor account handling to solve account logout issues

Half-Shot:
This PR does a lot, but unfortunately it's mostly intertwined and had to be done in a block. I'll try to explain the situation as it is now:

Currently, users run through the oauth flow and we store the resulting access token / slack id and team id into the "accounts" json object inside a `MatrixUser`, which is stored as a JSON blob in the database. A holdover from the NeDB days. This made it difficult to run queries to tell which users were connected to which slack team.

This became a problem for account logout / removing teams. There isn't a good way to tell if a team is unused, or the user is the last account holder for that team because iterating over all those JSON blobs is difficult. We deferred handling this, and simply deleted the account from the user object so they couldn't see their account, but the team would still be present in the database as we couldn't tell if the user was the last connected user for that team.

Fast forward to today where the team count is used to decide if a customer can bridge their team into their paid bridge, and suddenly those leftover teams matter. This change makes it so that we store the linked accounts in a separate table for easy querying, and that we properly clean up leftover teams and accounts so the team count matches up properly.

### DIFF
--- a/src/Provisioning.ts
+++ b/src/Provisioning.ts
@@ -88,15 +88,9 @@ export class Provisioner {
     }
 
     private async determineSlackIdForRequest(matrixUserId, teamId) {
-        const matrixUser = await this.main.datastore.getMatrixUser(matrixUserId);
-        if (!matrixUser) {
-            // No Slack user entry found for MXID
-            return null;
-        }
-        const accounts: {[userId: string]: {team_id: string}} = matrixUser.get("accounts");
-        for (const [key, value] of Object.entries(accounts)) {
-            if (value.team_id === teamId) {
-                return key;
+        for (const account of await this.main.datastore.getAccountsForMatrixUser(matrixUserId)) {
+            if (account.teamId === teamId) {
+                return account.slackId;
             }
         }
         return null;
@@ -146,20 +140,15 @@ export class Provisioner {
     }
 
     @command("user_id", "slack_id")
-    private async logout(req, res, userId, slackId) {
+    private async logout(req: Request, res: Response, userId: string, slackId: string) {
         if (!this.main.oauth2) {
             res.status(HTTP_CODES.NOT_FOUND).json({
                 error: "OAuth2 not configured on this bridge",
             });
             return;
         }
-        let matrixUser = await this.main.datastore.getMatrixUser(userId);
-        matrixUser = matrixUser ? matrixUser : new MatrixUser(userId);
-        const accounts = matrixUser.get("accounts") || {};
-        delete accounts[slackId];
-        matrixUser.set("accounts", accounts);
-        await this.main.datastore.storeMatrixUser(matrixUser);
-        log.info(`Removed account ${slackId} from ${slackId}`);
+        const logoutResult = await this.main.logoutAccount(userId, slackId);
+        res.json({logoutResult});
     }
 
     @command("user_id", "team_id")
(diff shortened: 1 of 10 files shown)


src/Provisioning.ts
```diff
@@ -208,16 +197,14 @@ export class Provisioner {
...
             res.status(HTTP_CODES.NOT_FOUND).json({error: "User has no accounts setup"});
             return;
         }
-        const accounts = matrixUser.get("accounts");
-        const results = await Promise.all(Object.keys(accounts).map(async (slackId) => {
-            const account = accounts[slackId];
-            return this.main.datastore.getTeam(account.team_id).then(
-                (team) => ({team, slack_id: slackId}),
+        const results = await Promise.all(accounts.map(async (account) => {
+            return this.main.datastore.getTeam(account.teamId).then(
```
jaller94, line 206: async function, but it uses `.then()`
Would be easier to react with await.

src/datastore/postgres/schema/v6.ts
```diff
@@ -0,0 +1,34 @@
...
+        }
+        // tslint:disable-next-line: no-any
+        for (const [slack_id, account] of Object.entries<any>(user.get("accounts") || { })) {
+            await db.none("INSERT INTO linked_accounts VALUES (${user_id}, ${slack_id}, ${team_id}, ${team_name})", {
+                user_id: userData.userid,
+                slack_id,
+                team_id: account.team_id,
+                team_name: account.team_name,
+            });
+        }
```
erdnaxeli, line 32: You could do all this in only [one insert](https://stackoverflow.com/a/37302557) :

```suggestion
    const pgp = require('pg-promise')()  # tihs line is probably wrong
    cons cs = new pgp.helpers.ColumnSet(['user_id', 'slack_id', 'team_id', 'team_name'], {table: 'linked_accounts'};
    let values = [];

    for (const userData of users) {
        const user = new MatrixUser(userData.userid, JSON.parse(userData.json));
        if (!user.get("accounts")) {
            continue;
        }
        // tslint:disable-next-line: no-any
        for (const [slack_id, account] of Object.entries<any>(user.get("accounts") || { })) {
            values.push(
                {
                    user_id: userData.userid,
                    slack_id,
                    team_id: account.team_id,
                    team_name: account.team_name,
                }
            );
        }
    }
    
    const query = pgp.helpers.insert(values, cs);
    await db.none(query);
```

Disclaimer: I am definitively not a typescript dev :p

src/datastore/postgres/schema/v6.ts
```diff
@@ -0,0 +1,34 @@
...
+import { MatrixUser } from "matrix-appservice-bridge";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Create database
+    await db.none(`
+        CREATE TABLE linked_accounts (
+            user_id TEXT NOT NULL,
+            slack_id TEXT NOT NULL,
+            team_id TEXT NOT NULL,
```
erdnaxeli, line 11: don't you want this to be a foreign key to `teams.id`?

Half-Shot:
@erdnaxeli 
> If it would have been JSON or better JSONB, you could use the value inside it to do queries.

Believe it or not I actually tried a few times to write a query for this, but it make me sad and I figured I should probably try to shove this data into a database anyway at some point.

src/datastore/postgres/schema/v6.ts
Half-Shot: We tend to avoid foreign keys, at the moment. It's technically valid to have a linked_account entry with no associated team entry. If we run into issues though, you can blame me :p

src/datastore/postgres/schema/v6.ts
Half-Shot: Nice find, I am doing this one :)

PR #420, opened 2020-05-27
Fix puppets being constrained by team / matrix id

Half-Shot:
We accidentally made it so that two puppets could never be registered against the same mxid, or the same team_id. Let's fix that. Using `v7` because `v6` is part of #419

### DIFF
--- a/src/datastore/postgres/schema/v7.ts
+++ b/src/datastore/postgres/schema/v7.ts
@@ -1,0 +1,10 @@
+import { IDatabase } from "pg-promise";
+import { MatrixUser } from "matrix-appservice-bridge";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Drop constraints
+    await db.none(`
+        ALTER TABLE puppets DROP CONSTRAINT puppets_slackteam_key;
+        ALTER TABLE puppets DROP CONSTRAINT puppets_matrixuser_key;`);
+}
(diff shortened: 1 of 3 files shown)


src/datastore/postgres/schema/v7.ts
```diff
@@ -0,0 +1,10 @@
+import { IDatabase } from "pg-promise";
+import { MatrixUser } from "matrix-appservice-bridge";
+
+// tslint:disable-next-line: no-any
+export async function runSchema(db: IDatabase<any>) {
+    // Drop constraints
+    await db.none(`
+        ALTER TABLE puppets DROP CONSTRAINT puppets_slackteam_key;
+        ALTER TABLE puppets DROP CONSTRAINT puppets_matrixuser_key;`);
```
jaller94, line 9: I can't find the code that created these constraints.

`cons_puppets_uniq` is the only constraint on the table `puppets`.
https://github.com/matrix-org/matrix-appservice-slack/blob/bf18f238a925e85d39c6b3e68d0ee154d978d3fd/src/datastore/postgres/schema/v2.ts#L10

Half-Shot: https://github.com/matrix-org/matrix-appservice-slack/blob/hs/fix-puppet-constraint/src/datastore/postgres/schema/v2.ts#L6-L8

`UNIQUE` is essentially a shorthand form for writing constraints. To remove `UNIQUE`, you need to drop it.

PR #421, opened 2020-05-27
Template HTML files for the oauth success/failure page

### LINKED ISSUE #412: Support prettier HTML document rendering for OAuth
Right now, we have some pretty basic success/failure pages hardcoded into the Oauth endpoint handlers. We should support custom HTML pages for things like modular/matrix.org branding. Ideally, even the defaults could look a bit prettier.

Half-Shot:
Fixes #412 

E.g.
image

Supports templating the HTML page you see when finishing the oauth flow. This PR also removes support for the


templates/oauth_result.html.njk
```diff
@@ -0,0 +1,68 @@
...
...rduHGjFTO/e0yLdpANMkUdrF+/3uSIrPr06WNSc0QhRKAOHDhgmUcBfsnAnj17rICbNGli0SK6VapU0cCBA7V06VKTGE6Dk7BOzZo1lTt3bnlOu4MHD9aOHTt09uxZnTp1Slu3bjUTIAiZM2dWpkyZVK9ePU2dOtXOw2ZTO1KaCGBt2OWcOXNSCJCFNm3apGw/nzx5Yo7F9cuXL7fPWBMSFHXt2rXtekCha9blfUJCgtUG52XJkkV16tTRpEmTzLUICP3D5yKGAP4PgMaNG1uq2ZN26dLF/JtG5F40o/j4eHXs2FEVKlQwl0HbSG/FihXW9NyL97t27TIXwxQgANEJEyaYHOk7FPJ/RsDZasmSJdWyZUutXLnSa8dH1pAKPcSTAEVO76ABpksGkBApBySdFSl8+fIlBRSFjFVisxQ6a0OADNAUPXdXRLpbt25W5FmzZvW/hFLXgCeBDh06WLRpPOifF7MODoO14vFIiAy0aNFC0dHRNkO5Jw24DBkkkxBIlxrw7AO4EI5BpChiosy8Q6qdXxPhVatWWcfFeZBG0aJFzR4peE8CBCYxMdGaGwHBiVwN+E1CnjZKH0hNICYmxorNtX4IoHVciMgTWbJA52WIozm5EeFXBBhTxo8f778i9swA0yYSypEjhxUxEnIZcASQEBmgYzNyYKFkAPulX7gMQMBJiO4bGhpq5oCE/FrEqWsAKeTMmdMiSxFTA2jVSQgCjAKuBhg3mHMY9ugPngS4V3JyshUxUqO2mIeoAUZ1v/UB/saYywNWIuu8nVmGRoQLuSLGhSAFKMZqJk1GBeRGcTsXggAuRLH27t1bJUqUsBkLouwntm/frkuXLvlOgMjev3/fhissc9iwYSYjbsTkyIxEH3A2yg4Nbx87dqxJgymUTI0aNcrAuj4AAbJBYLgvNkud8PAWoowkrOvzLAQwmhOLnThxwiLOQMaOC0DolI7qtoFIjlkHCVAL6B7/R+unT59OiSgEeFhL9tgokd158+ZZkNh7MECyrgtMmglgd8iDGxM95nT2sdyAzLx+/dqy5J5kM+fwNz5jLeqHg30wgfCc7wHntqusx7rICuC4D/dN/YQ83Z5K/NuHXGl9petjlaAm8L/LQFpJ/CEQsASC/huaoP+OLOi/pQz674mD/pv6oP9fiWD9b5WfSkWrKgYZxAsAAAAASUVORK5CYII=">
```
jaller94, line 4: uff.. this needs to be base64 inline, because we can't host files from the bridge?

IPFS fanboy note:
While possibly shitty because for the majority of users it depends on a centralised gateway, content-addressable IPFS-URLs may be an option.
https://gateway.pinata.cloud/ipfs/QmZwXKujr1UZPfVMqFS8ABLzJSaz8F2gE4Ky29JQthFGmw

templates/oauth_result.html.njk
```diff
@@ -0,0 +1,68 @@
...
+        {% if success %}
+        <p>Your account (<code>{{ userId }}</code>) has been successfully connected to the Slack bridge.</p>
+        <small>You can safely close this window</small>
+        {% else %}
+            {% if reason === "token-not-known" %}
+                <p>The bridge cannot handle this request because the token provided is not known.</p>
+                <small>This can happen if the bridge was recently restarted. Try running the oauth process again.</small>
+            {% elif reason === "limit-reached" %}
+                <p>You have reached the maximum number of connected teams on this bridge instance.</p>
+                <small>Please contact your bridge administrator</small>
```
jaller94, line 22: I don't expect people know who their bridge admin is.
I think we should have a config field for the admin to leave their free text contact, be it a name, a Matrix ID or whatever.

templates/oauth_result.html.njk
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
jaller94, line 1: Why are the html, head and body tags omitted?

src/OAuth2.ts
```diff
@@ -114,10 +124,16 @@ export class OAuth2 {
...
     }
 
-    private makeRedirectURL(roomOrString: string| BridgedRoom): string {
-        if (typeof roomOrString !== "string") {
-            roomOrString = roomOrString.InboundId;
-        }
-        return `${this.redirectPrefix}${roomOrString}/authorize`;
+    public getHTMLForResult(success: boolean, code: number, userId: string|null, reason?: "error"|"limit-reached"|"token-not-known") {
+        return this.oauthTemplate.render({
+            success,
```
jaller94, line 129: Can we name this `isSuccess` to indicate it's a boolean like `isPuppeting`?

templates/oauth_result.html.njk
Half-Shot: Hah, to be honest this was just to cut down on the number of moving parts. Serving files could be done, but felt like more effort than inlining some relatively small image files (ones SVG, the other is 48x48)

Half-Shot: We could use IPFS, but I just think it's the wrong place to introduce it. I really want this page to have zero dependencies and just work.

templates/oauth_result.html.njk
Half-Shot: I'm really on the fence about it. It feels like we could just tell admins to add their name to the template (as I think matrix.org would definitely add some custom content to this). 

Otoh, maybe customizing templates is more than most admins want to do.

templates/oauth_result.html.njk
Half-Shot: https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags

The general wisdom seems to be that if you are targeting HTML5 supporting browsers, you can and should omit the tags. I don't think we need to support IE since Riot doesn't, which makes this thing a bandwidth saver.

jaller94: ok... i guess.. nothing wrong
I'll try to ignore that these tips come from the company bringing us Chrome and AMP.

templates/oauth_result.html.njk
jaller94: good point, they can modify the templates.

templates/oauth_result.html.njk
Half-Shot: Hah, definitely healthy to have scepticism where Google is concerned but I think this is a safe suggestion :)

PR #422, opened 2020-05-27
Drop support for oauth with webhooks

Half-Shot:
This is actually implemented in #421 but I've split it into this PR to make reviewing that PR easier. We used to support authenticating via oauth to allow for richer bridging with webhooks, but to be honest it's a bit pointless when you can do the whole thing through oauth without having to setup cumbersome webhooks.

We actually stopped this from working ages ago, so this is just a cleanup mission.


Half-Shot:
apparently that happened via #421. Okay then :D

PR #423, opened 2020-05-28
Set profile on startup

Half-Shot:
To make the bot look a bit prettier, and to avoid that issue where joining over federation breaks. https://github.com/matrix-org/synapse/issues/2807

### DIFF
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -113,3 +113,7 @@ provisioning:
 puppeting:
   # Should the bridge allow users to puppet their accounts
   enabled: false
+
+bot_profile: # Set the bot's profile on startup, optional
+  displayname: Slack Bridge # Set the displayname, optional
+  avatar_url: "mxc://half-shot.uk/ea64c71ee946ca2f61379abefe2c7d977d276fbb" # Set the avatar, optional
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -767,7 +767,7 @@ export class Main {
     private async applyBotProfile() {
         log.info("Ensuring the bridge bot is registered");
         const intent = this.botIntent;
-        // The bot believes itelf to always be registered, even when it isn't.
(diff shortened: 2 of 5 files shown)


config/config.sample.yaml
```diff
@@ -112,4 +112,8 @@ provisioning:
 
 puppeting:
   # Should the bridge allow users to puppet their accounts
-  enabled: false
\ No newline at end of file
+  enabled: false
+
+bot_profile: # Set the bot's profile on startup, optional
+  displayname: Slack Bridge # Set the displayname, optional
+  avatar_url: "mxc://half-shot.uk/ea64c71ee946ca2f61379abefe2c7d977d276fbb" # Set the avatar, optional
```
jaller94, line 119: I'd comment out all of `bot_profile` as it's optional.
Do you want to keep your MXC URL in `avatar_url`?

src/Main.ts
```diff
@@ -848,6 +863,7 @@ export class Main {
             checkToken: false,
         });
 
+        await this.applyBotProfile();
```
jaller94, line 866: I'd like this in a try catch block, where an error outputs an error message but does not cause the bridge to crash.

src/Main.ts
```diff
@@ -764,6 +764,21 @@ export class Main {
         });
     }
 
+    private async applyBotProfile() {
+        log.info("Ensuring the bridge bot is registered");
+        const intent = this.botIntent;
+        // The bot believes itelf to always be registered, even when it isn't.
```
jaller94, line 770:
```suggestion
        // The bot believes itself to always be registered, even when it isn't.
```

config/config.sample.yaml
Half-Shot: Hmm, so I kept it in because in terms of defaults, I like the idea of people's bots having profiles when they set the bridge up. I left it as optional, because there will be cases where people don't want this, but I think it should be opt out.

PR #424, opened 2020-05-28
Fix puppeted edits

### LINKED ISSUE #350: M->S edits are broken if user is puppeted
It works only if user is unpuppeted.

Error message is:
```
ERROR Main Failed processing matrix edit:  { Error: An API error occurred: cant_update_message
    at Object.platformErrorFromResult (/opt/matrix-appservice-slack/node_modules/@slack/web-api/dist/errors.js:50:33)
    at WebClient.apiCall (/opt/matrix-appservice-slack/node_modules/@slack/web-api/dist/WebClient.js:382:28)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  code: 'slack_webapi_platform_error',
  data:
   { ok: false,
     error: 'cant_update_message',
     response_metadata: { scopes: [Array], acceptedScopes: [Array] } } }
```

Bridge version: 1.0.2

Half-Shot:
Fixes #350


src/BridgedRoom.ts
```diff
@@ -266,18 +278,18 @@ export class BridgedRoom {
...
+            return;
         }
+        // We must do this before sending to avoid racing
+        // Use the unicode key for uniqueness
+        this.addRecentSlackMessage(`reactadd:${relatesTo.key}:${client.id}:${event.slackTs}`);
 
         // TODO: This only works once from matrix if we are sending the event as the
         // bot user.
-        const res = await client.reactions.add({
+        const res = await client.client.reactions.add({
```
jaller94, line 292: `client.client.` is confusing.
I would prefer the outer wrapper client to be renamed to something that indicates what else the object does contain.
Of unwrap the object right away:
```
const { clientId: id, client } = getClientForRequest()
```

Half-Shot: Ack, I couldn't unwrap because it was null but I can fix the naming.

PR #425, opened 2020-05-28
Add /ready endpoint

Half-Shot:
So containers can notice when the bridge is ready and accepting messages


src/Main.ts
```diff
@@ -826,6 +828,27 @@ export class Main {
...
+        });
+
+        this.bridge.addAppServicePath({
+            handler: this.onReadyProbe.bind(this.bridge),
+            method: "GET",
+            path: "/ready",
+            checkToken: false,
+        });
+
+        this.stateStorage = new StateLookup({
```
jaller94, line 846: I don't understand what paul was warning against and why this does not apply anymore.

Half-Shot: It was about the state storage not being ready until after the bridge had started, but tbh I don't think it's been a problem for us in a long time.

PR #431, opened 2020-06-26
Don't over-redact events

### LINKED ISSUE #430: Slack bridge over-redacts a Matrix users redaction
This is likely because the bridge sees a matrix redaction, redacts it on slack, which causes it to redact on Matrix. We should fix this loop.

jaller94:
Fixes #430.


Half-Shot:
How is this one looking?

jaller94:
Rick asked me to not mess with the database but query the current state of the Matrix event before redacting it.
I don't know how to ask a Matrix server, if an event has been redacted by another event.

Half-Shot:
Ugh, short answer is you cannot. But since the loop comes from the slack side we could surely delete the event_id from the DB when we get a redaction from matrix so it won't be possible to over-redact?

jaller94:
That is what I had started to implement here, before I got told to try checking the redaction.
Will continue with the "delete the event from the DB" route on Monday.

jaller94:
## Current state
* `From Matrix; redacted on Slack` – "deleted by Slack Bridge" on Matrix
* `From Matrix; redacted on Matrix` – works
* `From Slack; redacted on Slack` – still visible on Matrix
* `From Slack; redacted on Matrix` – still visible on Slack; over-redacted on Matrix

Half-Shot:
Yeah, that's concerning.

`From Matrix; redacted on Slack – "deleted by Slack Bridge" on Matrix`  should really be mapping to the person who deleted it, not the bridge bot.

`From Slack; redacted on Slack – still visible on Matrix` is particularly problematic because that's an easy one in theory to solve.

`From Slack; redacted on Matrix – still visible on Slack; over-redacted on Matrix` is also bad because it obscures the redactor, also I would have expected deleting from the DB to solve this.

Half-Shot:
sounds like it's solved.

PR #433, opened 2020-07-17
Add optional display name suffix

dalcde:
This allows us to distinguish between Slack users and Matrix users. This
is especially useful when using Matrix as a bridging hub.

### DIFF
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -33,6 +33,7 @@ oauth2:
 
 # The prefix to give slack users 
 username_prefix: "slack_" # Required
+display_name_suffix: "[s]" # Optional
 
 homeserver: # Required
   # The domain name of your homeserver
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -594,7 +594,7 @@ export class BridgedRoom {
         this.addRecentSlackMessage(message.ts);
         try {
             const ghost = await this.main.ghostStore.getForSlackMessage(message, this.slackTeamId!);
-            await ghost.update(message, this);
+            await ghost.update(message, this, this.main.config);
             await ghost.cancelTyping(this.MatrixRoomId); // If they were typing, stop them from doing that.
             this.slackSendLock = this.slackSendLock.finally(async () => {
                 return this.handleSlackMessage(message, ghost, content);
@@ -619,7 +619,7 @@ export class BridgedRoom {
             return;
         }
         const ghost = await this.main.ghostStore.getForSlackMessage(message, teamId);
-        await ghost.update(message, this);
+        await ghost.update(message, this, this.main.config);
 
         const event = await this.main.datastore.getEventBySlackId(message.item.channel, message.item.ts);
 
--- a/src/SlackGhost.ts
+++ b/src/SlackGhost.ts
@@ -18,6 +18,7 @@ import { Logging, Intent } from "matrix-appservice-bridge";
 import * as rp from "request-promise-native";
 import * as Slackdown from "Slackdown";
 import { BridgedRoom } from "./BridgedRoom";
+import { IConfig } from "./IConfig";
 import { ISlackUser } from "./BaseSlackHandler";
 import { WebClient } from "@slack/web-api";
 import { BotsInfoResponse, UsersInfoResponse } from "./SlackResponses";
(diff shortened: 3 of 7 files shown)


src/BridgedRoom.ts
```diff
@@ -594,7 +594,7 @@ export class BridgedRoom {
         this.addRecentSlackMessage(message.ts);
         try {
             const ghost = await this.main.ghostStore.getForSlackMessage(message, this.slackTeamId!);
-            await ghost.update(message, this);
+            await ghost.update(message, this, this.main.config);
```
Half-Shot, line 597: There probably isn't much point passing in `this`, and also `this.main.config`

src/SlackGhost.ts
```diff
@@ -87,7 +88,7 @@ export class SlackGhost {
         };
     }
 
-    public async update(message: {user_id?: string, user?: string}, room: BridgedRoom) {
+    public async update(message: {user_id?: string, user?: string}, room: BridgedRoom, config: IConfig) {
```
Half-Shot, line 91: I recommend passing in the config in the constructor, and storing the suffix in a property rather than passing it into the update method.

src/SlackGhost.ts
```diff
@@ -115,15 +116,19 @@ export class SlackGhost {
...
         if (!slackUser.profile) {
             return;
         }
         let changed = false;
-        if (slackUser.profile.display_name && this.displayName !== slackUser.profile.display_name) {
-            await this.intent.setDisplayName(slackUser.profile.display_name);
-            this.displayname = slackUser.profile.display_name;
-            changed = true;
+        if (slackUser.profile.display_name) {
+            const newDisplayname = slackUser.profile.display_name + config.display_name_suffix
```
Half-Shot, line 125: Ideally please use template syntax
```suggestion
            const newDisplayname = `${slackUser.profile.display_name}${config.display_name_suffix}`
```

config/config.sample.yaml
```diff
@@ -33,6 +33,7 @@ oauth2:
```
Half-Shot, line 33: You also need to create a changelog file as per https://github.com/matrix-org/matrix-appservice-slack/blob/develop/CONTRIBUTING.md#making-pull-requests

Half-Shot: Also, please sign off :)

src/BridgedRoom.ts
dalcde: `main` is private, so we can't access `this.main.config`

src/SlackGhost.ts
dalcde: That was my original strategy but it became messy when I had to incorporate the tests. I could try doing it again though

config/config.sample.yaml
dalcde: Done!

src/SlackGhost.ts
jaller94: If we want to enforce this style, we should change the tslint.json.
Our current config:
```
    "prefer-template": [
      true,
      "allow-single-concat"
    ],
```
It's auto-fixable.
https://eslint.org/docs/rules/prefer-template

src/SlackGhost.ts
jaller94: In my opinion it's a good thing to not have the config on the SlackGhost as long as only `update` uses it.

jaller94:
@dalcde The PR's description makes me wonder if you know of the config `username_prefix`.
It checks your Acceptance Criteria "This allows us to distinguish between Slack users and Matrix users.".

dalcde:
On Wed, Jul 22, 2020 at 02:26:29AM -0700, Christian Paul wrote:
>@dalcde The PR's description makes me wonder if you know of the config `username_prefix`.
>It checks your Acceptance Criteria "This allows us to distinguish between Slack users and Matrix users.".

I am bridging matrix to other platforms. Since the local part that uses 
the username_prefix is opaque, it is not suitable as a display name in 
other platforms, so I must use the display name.

jaller94:
Yep, I mistook display name for username (commonly called "localpart" in Synapse). My bad.

src/SlackGhost.ts
Half-Shot: Yeah, let's do that in another PR!

src/BridgedRoom.ts
Half-Shot: Ah, how annoying. Fair doos.

src/SlackGhost.ts
Half-Shot: Would still be nice to pass the suffix into the constructor then rather than the update method, imo

Half-Shot:
@dalcde Hiya, sorry this has been left to rot for two years. How would you feel about rebasing and finish off the changes?

dalcde:
I am no longer using this, but other people are free to pick it up and finish it. Otherwise feel free to close the PR

PR #445, opened 2020-07-23
Fixes the admin command "help oauth" and better types for AdminCommands

jaller94:
Includes #443 

## Before
`help foo` searched for a method `foo` on AdminCommands. `help oauth` was broken, because the method has a different name.
broken oauth

But other class methods could be called, even though they are not admin commands.
Other examples like this: `help yargs`, `help commands`, `help constructor`, `help __prototype__`. 😐️
I haven't found any security issues or ways to crash the bridge with to this, however, I think this PR mitigates the risk that there is a way to exploit this.
broken parse

## After
fixed oauth

fixed parse

### DIFF
--- a/src/AdminCommands.ts
+++ b/src/AdminCommands.ts
@@ -306,7 +306,7 @@ export class AdminCommands {
                     respond("Oauth is not configured on this bridge");
                     return;
                 }
-                const token = this.main.oauth2.getPreauthToken(userId);
+                const token = this.main.oauth2.getPreauthToken(userId!);
                 const authUri = this.main.oauth2.makeAuthorizeURL(
                     token,
                     token,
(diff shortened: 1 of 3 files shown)


src/AdminCommands.ts
```diff
@@ -304,9 +331,12 @@ export class AdminCommands {
...
             "help [command]",
             "describes the commands available",
-            ({respond, command}) => {
+            ({respond, command}: {
+                respond: ResponseCallback,
+                command?: string,
+            }) => {
                 if (command) {
-                    const cmd = this[command as string] as AdminCommand;
+                    const cmd = this.commands.find((adminCommand) => (adminCommand.command.split(' ')[0] === command));
```
jaller94, line 339: This is the line that fixes the `help oauth` and protects against calling `.detailedDescription()` on other class members.

PR #448, opened 2020-07-24
Do not bridge deleted users on startup

### LINKED ISSUE #447: Team sync can create deleted users, even if they never existed on the bridge.

Half-Shot:
Fixes #447

### DIFF
--- a/src/TeamSyncer.ts
+++ b/src/TeamSyncer.ts
@@ -128,7 +128,7 @@ export class TeamSyncer {
                 queue.add(() => this.syncChannel(teamId, item));
             } else {
                 // tslint:disable-next-line: no-floating-promises
-                queue.add(() => this.syncUser(teamId, team.domain, item, true));
+                queue.add(() => this.syncUser(teamId, team.domain, item));
             }
         }
         await queue.onIdle();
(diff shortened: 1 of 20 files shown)


Half-Shot:
They'll linger forever as a ghost. I guess the right solution would be to check our DB to see if we've created them a user before?

src/TeamSyncer.ts
```diff
@@ -207,13 +207,17 @@ export class TeamSyncer {
         }
     }
 
-    public async syncUser(teamId: string, domain: string, item: ISlackUser) {
+    public async syncUser(teamId: string, domain: string, item: ISlackUser, initial = false) {
         log.info(`Syncing user ${teamId} ${item.id}`);
         const slackGhost = await this.main.ghostStore.get(item.id, domain, teamId);
```
jaller94, line 212: This function creates a ghost, if none does exist.
Do we need to check `item.deleted === true && initial` before calling `ghostStore.get()` to avoid creating a ghost that we don't want?

Half-Shot:
I managed to confuse this one with #450, can we merge that one first? :)

src/TeamSyncer.ts
```diff
@@ -209,7 +209,12 @@ export class TeamSyncer {
 
     public async syncUser(teamId: string, domain: string, item: ISlackUser) {
         log.info(`Syncing user ${teamId} ${item.id}`);
-        const slackGhost = await this.main.ghostStore.get(item.id, domain, teamId);
+        const existing = await this.main.ghostStore.getExisting(this.main.ghostStore.getUserId(item.id, domain));
+        if (item.deleted && !existing) {
+            // This is a deleted user that we've never seen, bail.
+            return;
+        }
+        const slackGhost = existing || await this.main.ghostStore.get(item.id, domain, teamId);
```
jaller94, line 217: `existing` made me thing it was a boolean, until I looked it up.
```suggestion
        const existingGhost = await this.main.ghostStore.getExisting(this.main.ghostStore.getUserId(item.id, domain));
        if (item.deleted && !existingGhost) {
            // This is a deleted user that we've never seen, bail.
            return;
        }
        const slackGhost = existingGhost || await this.main.ghostStore.get(item.id, domain, teamId);
```

PR #449, opened 2020-07-24
Sync channel membership after connecting room

Half-Shot:
It's a fairly each change to make the experience better :)

As @Cadair requested, this now syncs joins and leaves incrementally as well as ensures they sync on startup.

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1114,11 +1114,6 @@ export class Main {
             }
             room.setBotClient(slackClient);
             room.SlackChannelName = infoRes.channel.name;
-
-            // Perform syncing asyncronously.
-            this.teamSyncer?.syncMembershipForRoom(matrixRoomId, opts.slack_channel_id, opts.team_id, slackClient).catch((ex) => {
-                log.warn(`Failed to sync membership for ${opts.slack_channel_id}:`, ex);
-            });
         }
 
         if (isNew) {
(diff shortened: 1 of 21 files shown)


src/Main.ts
```diff
@@ -1114,6 +1114,11 @@ export class Main {
             }
             room.setBotClient(slackClient);
             room.SlackChannelName = infoRes.channel.name;
+
+            // Perform syncing asyncronously.
```
jaller94, line 1118:
```suggestion
            // Perform syncing asynchronously.
```

src/Main.ts
```diff
@@ -1114,6 +1114,11 @@ export class Main {
...
+            this.teamSyncer?.syncMembershipForRoom(matrixRoomId, opts.slack_channel_id, opts.team_id, slackClient).catch((ex) => {
```
jaller94, line 1119: Do you not want to await the sync, so it doesn't race with the code below?

Cadair:
What about also doing this for all existing channels? :angel:

Half-Shot:
>What about also doing this for all existing channels? angel

This should be done on startup. `TeamSyncer.syncAllTeams` is called, which calls `syncItems(teamId, client, "channel");` which calls `syncChannel`. We also sync membership events from slack as they happen.

src/Main.ts
Half-Shot: Moved this to the end.

src/TeamSyncer.ts
```diff
@@ -208,11 +210,22 @@ export class TeamSyncer {
...
-    public async syncUser(teamId: string, domain: string, item: ISlackUser) {
+    public async syncUser(teamId: string, domain: string, item: ISlackUser, newUser = false) {
         log.info(`Syncing user ${teamId} ${item.id}`);
         const slackGhost = await this.main.ghostStore.get(item.id, domain, teamId);
         if (item.deleted !== true) {
             await slackGhost.updateFromISlackUser(item);
+            for (const teamRoom of this.main.rooms.getBySlackTeamId(teamId)) {
+                if (teamRoom.IsPrivate || teamRoom.SlackType !== "channel") {
+                    // We only want PUBLIC rooms
+                    return;
```
jaller94, line 221: I think you meant to skip of this one room, but not terminate the sync for the user.
```suggestion
                    continue;
```

Issue #432, opened 2020-07-16
PNG images are sent with `m.file` msgtype instead of `m.image`, causing failure to generate preview


williamkray:
PNG image files uploaded via normal matrix clients behave properly because their msgtype is `m.image`. Those that are created by Slack users are uploaded as type `m.file` and so matrix clients only prompt users to download the file, without showing any sort of preview of the image.

JPG image files use the appropriate msgtype and render correctly. GIF images do not, and behave similarly to png.

williamkray:
the more i look at this, reviewing the source, the less it made sense... so i just tried using a very small png image, and the message/image upload was sent with the proper msgtype. so perhapse there's some sort of file processing issue because of the size or amount of time it takes.

williamkray:
further testing makes this look sporadic, but somehow associated with filesize or processing time. here's an example of an incorrectly identified message upload:

```{
  "type": "m.room.message",
  "sender": "@slack_jobmachine_U03MCE1DE:jobmachine.org",
  "content": {
    "body": "external-content.duckduckgo.com.jpg",
    "info": {
      "mimetype": "",
      "size": 399412
    },
    "msgtype": "m.file",
    "url": "mxc://jobmachine.org/yVMqIZUrkqgBjxvAjWiisSbO"
  },
  "origin_server_ts": 1594932734623,
  "unsigned": {
    "age": 257
  },
  "event_id": "$va593EMssZzlMI6iMfbeU82W7WFPHBQKkAmm0wqhGMw",
  "room_id": "!nwbqPRuaJbSnQTzCmo:jobmachine.org"
}
```

auscompgeek:
The `"mimetype": ""` seems rather suspicious here. Wonder when that'd be empty string.

williamkray:
yeah i'm having a hard time finding consistency... it apparently can happen with all filetypes, but could be I/O related. if someone posts two pictures, one right after the other, i've seen the first one come through just fine, and the second one doesn't (missing mimetype and wrong msgtype) even if it's a significantly smaller filesize.

seems like either way, the process is running into some sort of race condition wherein it can't identify the mimetype of the file, and therefore just dumps an m.file message instead of the appropriate version.

[...]

PR #450, opened 2020-07-24
Fetch slack file info if we are missing a mimetype

Half-Shot:
As a workaround for #432

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -656,7 +656,7 @@ export class BridgedRoom {
         sneakyThis[key] = value;
         this.dirty = true;
     }
-    
+
     private async handleSlackMessageFile(file: ISlackFile, slackEventId: string, ghost: SlackGhost) {
         const maxUploadSize = this.main.config.homeserver.max_upload_size;
         const channelId = this.slackChannelId;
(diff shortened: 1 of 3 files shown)


src/BridgedRoom.ts
```diff
@@ -657,6 +657,102 @@ export class BridgedRoom {
...
+                    uri: file.url_private!,
+                });
+            } catch (ex) {
+                log.error("Failed to download snippet", ex);
+                return;
+            }
+            let htmlCode = "";
+            // Because escaping 6 backticks is not good for readability.
+            // tslint:disable-next-line: prefer-template
+            const code = "```" + `\n${htmlString}\n` + "```";
```
jaller94, line 684: Is there a way to escape ``` if it was part of htmlString?

src/BridgedRoom.ts
```diff
@@ -657,6 +657,102 @@ export class BridgedRoom {
...
+            htmlCode += substitutions.htmlEscape(htmlString);
+            htmlCode += "</code></pre>";
+
+            const messageContent = {
+                body: code,
+                format: "org.matrix.custom.html",
+                formatted_body: htmlCode,
+                msgtype: "m.text",
+            };
+            await ghost.sendMessage(this.matrixRoomId, messageContent, channelId, slackEventId);
```
jaller94, line 699: We don't seem to check or handle a possible rejection due to the snippet possibly being too long.
I'm not asking you to implement this now, but file it, if you think that's a valid feature request that will come up.

Half-Shot: Good point, we should upload as a file if it's more than >~64Kb

Half-Shot: Created issue #450 to track.

src/BridgedRoom.ts
Half-Shot: Opened #461

src/BridgedRoom.ts
Half-Shot: Opened #460

PR #451, opened 2020-07-25
Update PgDatastore.ts

vitaly-t:
Proper way of generating multi-row inserts. See also - [Multi-row insert with pg-promise](https://stackoverflow.com/questions/37300997/multi-row-insert-with-pg-promise).

Your current implementation is wrong on many levels.

This will solve about 90% of problems, creating room for further improvements.

### DIFF
--- a/src/datastore/postgres/PgDatastore.ts
+++ b/src/datastore/postgres/PgDatastore.ts
@@ -209,7 +209,7 @@ export class PgDatastore implements Datastore {
             status: entry.status,
             user_id: entry.user_id,
         };
-        const statement = PgDatastore.BuildUpsertStatement("teams", "id", props);
+        const statement = PgDatastore.BuildUpsertStatement("teams", ["id"], [props]);
         await this.postgresDb.none(statement, props);
     }
 
@@ -316,7 +316,7 @@ export class PgDatastore implements Datastore {
     }
 
     public async setUserAdminRoom(matrixuser: string, roomid: string): Promise<null> {
-        const statement = PgDatastore.BuildUpsertStatement("user_admin_rooms", "matrixuser", {matrixuser, roomid});
+        const statement = PgDatastore.BuildUpsertStatement("user_admin_rooms", ["matrixuser"], [{matrixuser, roomid}]);
         return this.postgresDb.none(statement, {matrixuser, roomid});
     }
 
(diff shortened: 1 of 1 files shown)


src/datastore/postgres/PgDatastore.ts
```diff
@@ -209,7 +209,7 @@ export class PgDatastore implements Datastore {
             status: entry.status,
             user_id: entry.user_id,
         };
-        const statement = PgDatastore.BuildUpsertStatement("teams", "id", props);
+        const statement = PgDatastore.BuildUpsertStatement("teams", ["id"], [props]);
         await this.postgresDb.none(statement, props);
```
jaller94, line 213: Does this still need to be called with `props` or are the values inserted in `pgp.helpers.insert(values, cs)`?

vitaly-t:
In relation to being `"wrong on many level"`...

* You are using string concatenation, which is a strict no-no, as it may result in SQL injection
* You are generating a lot of variables via strings, which makes it even more error-prone, and without possibility to re-cast anything, for example
* You should always work with arrays of objects, and not with just a single object, as such things do not scale into multi-row operations.
* When in conflict, data is supposed to be updated from `EXCLUDED` table, but your code instead assigns columns to themselves, which results into nothing.

And then some minor stuff... :smile:

src/datastore/postgres/PgDatastore.ts
vitaly-t: In your code, `props` is the actual data, so i'm not sure about your question, whether we can skip the actual data to be inserted?

vitaly-t: > why `conflictKeys` and `values` became arrays

Because they were changed to be generic, allowing the last parameter of `BuildUpsertStatement` to be an array objects, which will automatically generate a multi-row insert.

src/datastore/postgres/PgDatastore.ts
```diff
@@ -403,11 +403,9 @@ export class PgDatastore implements Datastore {
...
     }
 
-    private static BuildUpsertStatement(table: string, conflictKey: string, keyValues: {[key: string]: string}) {
-        const keys = Object.keys(keyValues).join(", ");
-        const keysValues = `\${${Object.keys(keyValues).join("}, ${")}}`;
-        // tslint:disable-next-line: prefer-template
-        const keysSets = Object.keys(keyValues).slice(1).map((k) => `${k} = \${${k}}`).join(", ");
-        return `INSERT INTO ${table} (${keys}) VALUES (${keysValues}) ON CONFLICT (${conflictKey}) DO UPDATE SET ${keysSets}`;
+    private static BuildUpsertStatement(table: string, conflictKeys: string[], values: Array<{[key: string]: any}>) {
+        const cs = new pgp.helpers.ColumnSet(values[0], {table});
```
jaller94, line 407: Is it a problem if `values[0]` contains less columns than `values[1]`?

src/datastore/postgres/PgDatastore.ts
jaller94: Looks redundant to me. If you don't remove it, I'll remove the 2nd argument to `.none` in a follow-up PR.
[insert](https://vitaly-t.github.io/pg-promise/helpers.html#.insert) does put the data in the statement while not generating placeholders.

src/datastore/postgres/PgDatastore.ts
vitaly-t: Yes, it is. I generally do not recommend use of an object to extract column names. That's what I meant by:

> This will solve about 90% of problems, creating room for further improvements.

A better approach is to pre-declare each `ColumnSet`, specifying column names statically.

vitaly-t: > Looks redundant to me. 

Not quite. You'd need to add static ColumnSet, and make use of it, then it will be redundant and can be refactored. In the current version however the code relies on column names within the data object.

vitaly-t: @jaller94 You will do well, simply following the logic shown in [Multi-row insert with pg-promise](https://stackoverflow.com/questions/37300997/multi-row-insert-with-pg-promise) ;)

Issue #452, opened 2020-07-26
Threads and reactions not working if message originated from slack


stemnic:
Using the latest develop branch which would be dcb7a8d3a82f4c37ec5f2bb3b976128857afc0fd

I'm having a problem where bridge channels halfway work. Messages are passed between them but threads do not work if the message originated from slack.

Reactions seems also to be limited in this way that the bridge is only able to react to its own posts from the matrix side.

This specific channel is linked with the team_sync feature however this is also an issue if you manually link the channel.

Regarding configuration, I think I am using a combination RTM and events (documentation is a little unclear on how this should be setup)

Would appreciate any input on what I might be doing wrong :)

## thread start from matrix
![threadstartmatrix(matrix)](https://user-images.githubusercontent.com/1348006/88477050-1b86df80-cf3d-11ea-80bd-f1c0704c7d16.png)
![threadstartmatrix(slack)](https://user-images.githubusercontent.com/1348006/88477051-1c1f7600-cf3d-11ea-83be-fb0217260c7f.png)

## thread start from slack
![threadstartslack(slack)](https://user-images.githubusercontent.com/1348006/88477053-1c1f7600-cf3d-11ea-903d-00ec0ccd74a5.png)
![threadstartslack(matrix)](https://user-images.githubusercontent.com/1348006/88477052-1c1f7600-cf3d-11ea-92ff-38e7fed5fad3.png)

## Configuration

![event](https://user-images.githubusercontent.com/1348006/88477134-efb82980-cf3d-11ea-93fb-b4befe5b87bc.png)
![scope](https://user-images.githubusercontent.com/1348006/88477136-f050c000-cf3d-11ea-8986-4b737aa5bfbf.png)

```yaml
# Port for incoming slack requests from webhooks and event API messages.
slack_hook_port: 9898
# Prefix of incoming requests to strip when handing. This is NOT the bind host
  #NOTE: Unlike most of the other urls, this one cannot use localhost,
  #as this one must be publicly visible to the slack API
inbound_uri_prefix: "https://matrix.(domain):9898/"

# Both of the above options are optional if using RTM (see below), required otherwise.

# TLS files to be used when handing slack requests
# Optional
tls:
  key_file: /etc/letsencrypt/live/matrix.(domain)/privkey.pem
  crt_file: /etc/letsencrypt/live/matrix.(domain)/fullchain.pem

# Optional if slack_hook_port and inbound_uri_prefix are defined, required otherwise.
rtm:
[...]

PR #459, opened 2020-08-12
Bump matrix-appservice-bridge@1.13.2

Half-Shot:
in 1.13.1 we were returning undefined for `sendEvent` which was causing issues inserting events into the DB, in turn causing #452


Half-Shot:
(the linting was fixed in another PR)

PR #474, opened 2020-08-26
schema-fixes for direct_messages

### LINKED ISSUE #473: Direct message ADL schema / config does not match actual functionality
- [We allow "allow" and "deny" in the code](https://github.com/matrix-org/matrix-appservice-slack/blob/develop/config/config.sample.yaml#L119)
- [these](https://github.com/matrix-org/matrix-appservice-slack/blob/develop/config/slack-config-schema.yaml#L133-L149) should be arrays of strings, not strings

Half-Shot:
Fixes #473

### DIFF
--- a/changelog.d/474.bugfix
+++ b/changelog.d/474.bugfix
@@ -1,0 +1,1 @@
+Fix schema for puppeting.direct_messages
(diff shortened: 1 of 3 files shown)


jaller94 (approved):
Looks good to me.

PR #475, opened 2020-08-26
Sync users and channels on new team setup

Half-Shot:
We currently sync users and channels on startup, but this less useful if you've just added your team.

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1221,7 +1221,7 @@ export class Main {
                     "user",
                 );
             } catch (ex) {
-                log.info("Failed to sync members", ex);
+                log.warn("Failed to sync members", ex);
             }
 
             try {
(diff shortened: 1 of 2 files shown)


src/Main.ts
```diff
@@ -1211,6 +1212,28 @@ export class Main {
...
+            }
+
+            try {
+                await this.teamSyncer.syncItems(
+                    teamId,
+                    await this.clientFactory.getTeamClient(teamId),
+                    "channel",
+                );
+            } catch (ex) {
+                log.info("Failed to sync channels", ex);
```
jaller94, line 1234: I think this should be a warning.
```suggestion
                log.warn("Failed to sync members", ex);
            }

            try {
                await this.teamSyncer.syncItems(
                    teamId,
                    await this.clientFactory.getTeamClient(teamId),
                    "channel",
                );
            } catch (ex) {
                log.warn("Failed to sync channels", ex);
```

jaller94 (approved):
Looks good to me.

PR #476, opened 2020-09-01
Add configuration to deny slack channels by name or ID

Half-Shot:
This allows us to A/D channels from being bridged across on startup or through the provisioner based on name or id.

### DIFF
--- a/src/TeamSyncer.ts
+++ b/src/TeamSyncer.ts
@@ -22,6 +22,7 @@ import { ConversationsInfoResponse, UsersInfoResponse, ConversationsListResponse
 import { WebClient } from "@slack/web-api";
 import PQueue from "p-queue";
 import { ISlackUser } from "./BaseSlackHandler";
+import { DenyReason } from "./AllowDenyList";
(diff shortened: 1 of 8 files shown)


src/TeamSyncer.ts
```diff
@@ -135,13 +136,19 @@ export class TeamSyncer {
...
             return false;
         }
-        if (item === "channel" && (!teamConfig.channels || !teamConfig.channels.enabled)) {
-            return false;
+        if (item === "channel") {
+            if (!teamConfig.channels?.enabled) {
+                return false;
+            }
+            if (teamConfig.channels.allow_private === false) {
+                // Default to true.
```
jaller94, line 149: What's this "Default to true" comment about?
If the condition is true, `false` is returned.

jaller94: Ah, because of `=== false` where `undefined` would lead to the condition being false.

jaller94 (approved):
Looks good to me.
giphub

PR #477, opened 2020-09-01
Subscribe to additional events on the RTM API

Half-Shot:
This ensures we still create and remove rooms for channels in realtime rather than during startup.


jaller94 (approved):
Looks good to me.

PR #478, opened 2020-09-02
Do not reflect the bridge bot to Matrix

Half-Shot:
When the slack side bot joins the room, it gets reflected to Matrix. Since we never expect the bot to "speak" in the conversation, it's presence in the room is unwanted.


jaller94 (approved):
Nice! I always wondered how to remove it.

PR #484, opened 2020-09-08
Upgrade dependencies

jaller94:
All of these seem safe and I tested my Slack bridge a bit with the new dependencies installed.
`yargs` has breaking changes for invalid configs that don't affect us, because the repo has full control over the config and doesn't allow user input.


changelog.d/484.misc
```diff
@@ -0,0 +1 @@
+Upgrade dependencies with no affecting breaking changes
```
Half-Shot, line 1:
```suggestion
Upgrade dependencies
```

I wouldn't include the breaking changes line, we are expected to uphold that promise implicitly unless we bump a major version :)

Half-Shot (approved):
Otherwise, good to go!

changelog.d/484.misc
jaller94: ok

jaller94: Missed to change this before merging:
Changed in 944033902033c9f7a516afc9aa5b34cfeb89e640.

PR #485, opened 2020-09-09
Support removing reactions

### LINKED ISSUE #154: Implement removal of reactions
Currently we add reactions in both directions but can't remove them across the bridge.

For matrix -> slack reaction removal, the issue is when we *add* the reaction to slack, the `reactions.add` slack API method doesn't return the ts of the reaction event. You would have to listen for the event to come in on the event stream.

For slack -> matrix removal, the issue is that the slack event `reaction_removed` also dosen't give you the ts of the reaction event, it gives you the ts of the parent event as well as the sender of the reaction and the key of the reaction. This therefore means we would have to use the not-finished-yet aggregations API to query synapse for the event ID that matched those parameters.

jaller94:
Fixes #154 

Bridges reactions being removed. We can't use the APIs of Matrix nor Slack to identify bridged reactions.
So we store all reactions in a new table. This also means that we can only remove reactions we have bridged before.


src/SlackEventHandler.ts
```diff
@@ -324,24 +313,32 @@ export class SlackEventHandler extends BaseSlackHandler {
...
-        });
+            user_id: userOrBotId,
+        };
 
         if (event.type === "reaction_added") {
-            return room.onSlackReactionAdded(msg, teamId);
+            await room.onSlackReactionAdded(msg, teamId);
+            return;
+        } else if (event.type === "reaction_removed") {
+            // TODO Avoid processing our own events.
```
jaller94, line 333: Still TODO

src/SlackEventHandler.ts
```diff
@@ -231,26 +233,13 @@ export class SlackEventHandler extends BaseSlackHandler {
         // Only count received messages that aren't self-reflections
         this.main.incCounter(METRIC_RECEIVED_MESSAGE, {side: "remote"});
 
-        if (event.type === "channel_join") {
```
jaller94, line 234: May be needed for Slack RTM.

src/SlackEventHandler.ts
```diff
@@ -231,26 +233,13 @@ export class SlackEventHandler extends BaseSlackHandler {
...
-        const msg = Object.assign({}, event, {
+        const msg = {
+            ...event,
             channel_id: event.channel,
             team_domain: team.domain || team.id,
             team_id: teamId,
-            user_id: event.user || event.bot_id,
-        });
-
-        if (event.type === "reaction_added") {
```
jaller94, line 247: May be needed for Slack RTM.

changelog.d/485.feature
```diff
@@ -0,0 +1 @@
+Support removing reactions
```
Half-Shot, line 1:
```suggestion
Support removing reactions from Slack and Matrix messages
```

to give a bit more context

src/BridgedRoom.ts
```diff
@@ -292,22 +292,30 @@ export class BridgedRoom {
         this.addRecentSlackMessage(`reactadd:${relatesTo.key}:${id}:${event.slackTs}`);
 
         // TODO: This only works once from matrix if we are sending the event as the
-        // bot user.
+        // bot user. Search for #fix_reactions_as_bot.
```
Half-Shot, line 295: What does `Search for #fix_reactions_as_bot` mean?

src/BridgedRoom.ts
```diff
@@ -322,6 +330,20 @@ export class BridgedRoom {
 
         // If we don't get an event then exit
         if (event === null) {
+            const reactionEntry = await this.main.datastore.getReactionByMatrixId(message.room_id, message.redacts);
+
+            if (reactionEntry) {
+                log.debug(`Trying to redact reaction "${reactionEntry.reaction}" on message ${reactionEntry.slackMessageTs} in channel ${reactionEntry.slackChannelId}`);
```
Half-Shot, line 336: I'd make this an info, it's probably low noise and would help people debug.

src/BridgedRoom.ts
```diff
@@ -292,22 +292,30 @@ export class BridgedRoom {
...
         if (!res.ok) {
-            log.error("HTTP Error: ", res.error);
+            log.error(`HTTP Error from Slack when adding the reaction :${emojiKeyName}: to ${event.slackTs}: `, res.error);
             return;
         }
-        // TODO: Add this event to the event store
-        // Unfortunately reactions.add does not return the ts of the reactions event.
-        // So we can't store it in the event store
+
+        log.info(`Reaction :${emojiKeyName}: added to ${event.slackTs}`);
```
Half-Shot, line 308: I'd include the matrix event_id, room_id here too for extra debugability.

src/BridgedRoom.ts
```diff
@@ -640,8 +672,44 @@ export class BridgedRoom {
             return;
         }
         log.debug(`Sending reaction ${reactionKey} for ${event.eventId} as ${ghost.userId}`);
```
Half-Shot, line 674: Would also make this info, and include the channel and ts

src/BridgedRoom.ts
```diff
@@ -640,8 +672,44 @@ export class BridgedRoom {
...
+    ) {
+        if (msg.user_id === this.team!.user_id) {
+            return;
+        }
+        const originalEvent = await this.main.datastore.getReactionBySlackId(msg.item.channel, msg.item.ts, msg.user_id, msg.reaction );
+        if (!originalEvent) {
+            throw Error('unknown_reaction');
+        }
+        const botClient = this.main.botIntent.getClient();
+        botClient.redactEvent(originalEvent.roomId, originalEvent.eventId);
```
Half-Shot, line 711: await?

src/Main.ts
```diff
@@ -834,18 +834,21 @@ export class Main {
...
             });
             await new Promise((resolve, reject) => {
-                teamDatastore.loadDatabase((err) => {
-                if (err) {
-                    reject(err);
-                    return;
-                }
-                resolve();
-            }); });
+                teamDatastore.loadDatabase(err => err ? reject(err) : resolve());
```
Half-Shot, line 837: nice :+1:

src/SlackEventHandler.ts
```diff
@@ -217,6 +222,7 @@ export class SlackEventHandler extends BaseSlackHandler {
     protected async handleMessageEvent(event: ISlackMessageEvent, teamId: string) {
         const room = this.main.rooms.getBySlackChannelId(event.channel) as BridgedRoom;
         const team = await this.main.datastore.getTeam(teamId);
+        const userOrBotId = event.user || event.bot_id;
```
Half-Shot, line 225: Hmm, careful here. A bot has both a user and bot_id, which are two different things. Why do we want to pass the bot_id here?

src/SlackEventHandler.ts
```diff
@@ -228,29 +234,23 @@ export class SlackEventHandler extends BaseSlackHandler {
...
-        });
-
-        if (event.type === "reaction_added") {
-            return room.onSlackReactionAdded(msg, teamId);
-        }
-        // TODO: We cannot remove reactions yet, see https://github.com/matrix-org/matrix-appservice-slack/issues/154
-        /* else if (params.event.type === "reaction_removed") {
-            return room.onSlackReactionRemoved(msg);
-        } */
+            user_id: userOrBotId!,
```
Half-Shot, line 252: Ah, I see why you created `userOrBotId`. I'm not sure if we should pass this value to `onSlackUserJoin|Left` though.

src/datastore/NedbDatastore.ts
```diff
@@ -17,17 +17,32 @@ import { BridgedRoom } from "../BridgedRoom";
 import { SlackGhost } from "../SlackGhost";
```
Half-Shot, line 17: As per what we said in sync. If the code is done and working then don't remove it, but if it's still not right we don't need to support NeDB.

src/tests/integration/NedbDatastoreTest.ts
```diff
@@ -21,15 +22,54 @@ import { doDatastoreTests } from "./SharedDatastoreTests";
 describe("NedbDatastore", () => {
```
Half-Shot, line 22: Woo! Tests

Half-Shot (changes requested):
This is looking really good, but I have a few concerns. Presuming this is tested and my concerns are satisfied, we can probably tick this.

src/BridgedRoom.ts
jaller94: I added this to several line to easier find them when the feature gets implemented.
As I expect this to be a feature we'll want to implement, I wanted to add these bookmarks.

src/BridgedRoom.ts
jaller94: I'll instead add an info.warn if the API call fails.

src/datastore/NedbDatastore.ts
jaller94: All my manual testing was done with Postgres... there could be plain errors in this NeDB implementation.

src/BridgedRoom.ts
jaller94: Removed

src/BridgedRoom.ts
jaller94: Added that, made it info on success and warn on failure.

Half-Shot (approved):
I think there are a couple of leftover comments to address, but what I can see looks good.

src/BridgedRoom.ts
Half-Shot: @jaller94 so you don't miss this

PR #487, opened 2020-09-11
channel_join and channel_leave are subtypes

jaller94:
These are subtypes, so I would never expect them to equal `event.type`.

* https://api.slack.com/search?query=channel_join
* https://api.slack.com/events/message/channel_leave

I'm a bit scared of this potentially breaking things by enabling previously unused stuff.


Half-Shot:
FYI https://api.slack.com/changelog/2017-05-rethinking-channel-entrance-and-exit-events-and-messages

I think we should remove this bit of code altogether and use the events themselves.

Half-Shot (changes requested):
I'd remove this handler entirely as per the linked blog post.

Half-Shot (approved):
Works for me

PR #489, opened 2020-09-15
Remove support for removing reactions removal for NeDB

jaller94:
We haven't actually tested this and we don't have intentions to maintain this for the deprecated datastore NeDB.

If you want to use this (and other great features): Please use your bridge with PostgreSQL.


jaller94:
I had to disable `no-unused-expression` because the overrides are not actually working and I didn't see any evidence that tslint supports overrides.

Half-Shot (approved):
Looks agreeable to me.

PR #490, opened 2020-09-15
Migrate from TSLint to ESLint

jaller94:
TSLint got deprecated.
https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/FAQs.md

This moves to ESLint.


src/AdminCommands.ts
```diff
@@ -69,7 +71,7 @@ export class AdminCommands {
                 let nameFilter: RegExp;
 
                 if (team) {
-                    nameFilter = new RegExp(`^${quotemeta(team)}\.#`);
+                    nameFilter = new RegExp(`^${quotemeta(team)}.#`);
```
jaller94, line 74: This possibly should have been `\\.` to be a literal .

src/Provisioning.ts
```diff
@@ -31,11 +31,11 @@ type Param = string | { param: string, required: boolean};
 type Verbs = "getbotid"|"authurl"|"channels"|"getlink"|"link"|"logout"|"removeaccount"|"teams"|"accounts"|"unlink";
 
 // Decorator
-function command(...params: Param[]) {
```
Half-Shot, line 34: feels a big ugh here to have an arrow function for something that doesn't need access to the outside scope.

src/Provisioning.ts
```diff
@@ -203,9 +203,9 @@ export class Provisioner {
             return;
         }
         const results = await Promise.all(accounts.map(async (account) => {
-                const team = await this.main.datastore.getTeam(account.teamId)
-                return {team, slack_id: account.slackId};
-            })
+            const team = await this.main.datastore.getTeam(account.teamId)
```
Half-Shot, line 206: Surprised we don't enforce semi-colons? We should have a rule for that.

src/SlackClientFactory.ts
```diff
@@ -49,7 +49,7 @@ export class SlackClientFactory {
                     // non-ideal way to detect calls to slack.
                     webLog.debug.bind(webLog);
                     if (!this.onRemoteCall) { return; }
-                    const match = /apiCall\('([\w\.]+)'\) start/.exec(msg);
```
Half-Shot, line 52: Again, pretty sure this is meant to be a literal. We should double slash.

src/SlackGhost.ts
```diff
@@ -269,7 +269,7 @@ export class SlackGhost {
             return;
         }
 
-        const match = hash || avatarUrl.match(/\/([^\/]+)$/);
```
Half-Shot, line 272: We should be mindful here that this is a breaking change too?

src/TeamSyncer.ts
```diff
@@ -392,28 +392,23 @@ export class TeamSyncer {
                     text: `Hint: To bridge to Matrix, run the \`/invite @${user!.name}\` command in this channel.`,
                     channel: channelItem.id,
                 });
-            } catch (ex) {
+            } catch {
```
Half-Shot, line 395: Wonder if we should have logged the exception at debug level.

src/datastore/postgres/PgDatastore.ts
```diff
@@ -13,9 +13,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-// tslint thinks we can combine these two statements, but so far I have been unable to.
+// eslint thinks we can combine these two statements, but so far I have been unable to.
```
Half-Shot, line 16: does this work now?

src/substitutions.ts
```diff
@@ -29,9 +29,9 @@ const PILL_REGEX = /<a href="https:\/\/matrix\.to\/#\/(#|@|\+)([^"]+)">([^<]+)<\
  * Will return the emoji's name within ':'.
  * @param name The emoji's name.
  */
-export function getFallbackForMissingEmoji(name): string {
-    return `:${name}:`;
-}
+export const getFallbackForMissingEmoji = (name): string => (
```
Half-Shot, line 32: Again feels a little weird to enforce arrow functions wholesale. Makes more sense for nested functions, but not exported ones.

Half-Shot (changes requested):
Been meaning to get to this for a long time. There are a few questionable rule choices in here though.

src/AdminCommands.ts
```diff
@@ -369,6 +369,7 @@ export class AdminCommands {
             try {
                 let matched = false;
                 this.yargs.parse(argv, {
+                    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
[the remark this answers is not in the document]
jaller94: I think the rule is not cool with us using `foo ? a() : b()` to conditionally call functions instead of assigning the return value of the entire term.

.eslintrc.js
```diff
@@ -0,0 +1,173 @@
+module.exports = {
+    "root": true,
+    "env": {
+        "browser": true,
+        "es6": true,
```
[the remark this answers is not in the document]
jaller94: Seems still required as the default is ECMAScript 5.

.eslintrc.js
```diff
@@ -0,0 +1,173 @@
+module.exports = {
+    "root": true,
+    "env": {
+        "browser": true,
```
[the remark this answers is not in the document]
jaller94: Correct, I should have checked the automatically converted file.

src/SlackClientFactory.ts
jaller94: The dot character is always literal inside a `[]` list.

src/datastore/postgres/PgDatastore.ts
jaller94: Sadly, no.

src/substitutions.ts
jaller94: I didn't enable this but the automatic converter for tslint to eslint configs added this rule – we can remove it.

Usually, I try to avoid `function` in vanilla JS to avoid collisions in the global namespace.
I guess TypeScript makes everything a module anyway.

.eslintrc.js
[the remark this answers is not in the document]
jaller94: I removed this.
It's now a the same as in `-appservice-bridges`.

.eslintrc.js
```diff
@@ -0,0 +1,173 @@
...
+    ],
+    "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended"
+    ],
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "project": "tsconfig.json",
+        "sourceType": "module"
```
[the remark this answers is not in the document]
jaller94: I copied this from appservice-bridge while keeping the existing properties as `lint` crashed without them.

src/TeamSyncer.ts
jaller94: Added a `log.debug` of the error.

src/SlackGhost.ts
jaller94: `/` inside of `[]` is already a literal `/`

src/Provisioning.ts
jaller94: I set semi to "error" now.

src/Main.ts
```diff
@@ -1144,7 +1149,7 @@ export class Main {
                 room.SlackChannelName = channelInfo.channel.name;
             }
             isNew = true;
-            this.stateStorage?.trackRoom(matrixRoomId);
+            await this.stateStorage?.trackRoom(matrixRoomId);
```
jaller94, line 1152: This used to be a dangling promise that wasn't waited for.
I think it's meant to be waited for.

src/Main.ts
```diff
@@ -520,7 +522,7 @@ export class Main {
                 // Mark the room as active if we managed to join.
                 if (forRoom) {
                     forRoom.MatrixRoomActive = true;
-                    this.stateStorage?.trackRoom(ev.room_id);
+                    await this.stateStorage?.trackRoom(ev.room_id);
```
jaller94, line 525: This used to be a dangling promise that wasn't waited for.
I think it's meant to be waited for.

Half-Shot: Yeah, sounds good.

src/SlackClientFactory.ts
Half-Shot: oh, didn't know that.

Half-Shot (approved):
I think this looks generally good now. Ready to go from my side.

PR #499, opened 2020-09-22
Fix issue where threading would break

Half-Shot:
This boils down to awaiting a promise.

### DIFF
--- a/src/datastore/postgres/PgDatastore.ts
+++ b/src/datastore/postgres/PgDatastore.ts
@@ -153,22 +153,15 @@ export class PgDatastore implements Datastore {
 
     public async getEventBySlackId(slackChannel: string, slackTs: string): Promise<EventEntry|null> {
         log.debug(`getEventBySlackId: ${slackChannel} ${slackTs}`);
-        const events = await this.postgresDb.manyOrNone(
-            "SELECT * FROM events WHERE slackChannel = ${slackChannel} AND slackTs = ${slackTs}",
-            { slackChannel, slackTs }
-        );
-        if (events[0]) {
-            const e = events[0];
-            return {
+        return this.postgresDb.oneOrNone(
+            "SELECT * FROM events WHERE slackChannel = ${slackChannel} AND slackTs = ${slackTs} LIMIT 1",
+            { slackChannel, slackTs }, e => e && {
                 roomId: e.roomid,
                 eventId: e.eventid,
                 slackChannelId: slackChannel,
                 slackTs,
                 _extras: JSON.parse(e.extras),
-            };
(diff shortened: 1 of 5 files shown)


src/datastore/postgres/PgDatastore.ts
```diff
@@ -152,15 +152,23 @@ export class PgDatastore implements Datastore {
     }
 
     public async getEventBySlackId(slackChannel: string, slackTs: string): Promise<EventEntry|null> {
-        return this.postgresDb.oneOrNone(
+        log.debug(`getEventBySlackId: ${slackChannel} ${slackTs}`);
+        const events = await this.postgresDb.manyOrNone(
```
jaller94, line 156: To me the significant change isn't the await (which introduces no change as far as I can tell, but the change from `oneOrNone` to `manyOrNone`.
This does not seem to fix the root of the issue that we have one event stored multiple times in the database.

jaller94:
Newsfile is missing.

src/datastore/postgres/PgDatastore.ts
Half-Shot: So the problem really is that we have multiple matrix events for one slack event. The problem comes in because we end up with two results if we search by slackTs. This PR is more of an attempt to do a best-effort fetch the first event rather than catch-fire-and-fail.

Ideally, we wouldn't store multiple events.

src/datastore/postgres/PgDatastore.ts
```diff
@@ -152,15 +152,23 @@ export class PgDatastore implements Datastore {
...
                 roomId: e.roomid,
                 eventId: e.eventid,
                 slackChannelId: slackChannel,
                 slackTs,
                 _extras: JSON.parse(e.extras),
-            });
+            };
+        } else {
+            return null;
+        }
```
jaller94, line 171: /me makes a cheesy sales pitch about `LIMIT`.

```
        return this.postgresDb.oneOrNone(
            "SELECT * FROM events WHERE slackChannel = ${slackChannel} AND slackTs = ${slackTs} LIMIT 1",
            { slackChannel, slackTs }, e => e && {
                roomId: e.roomid,
                eventId: e.eventid,
                slackChannelId: slackChannel,
                slackTs,
                _extras: JSON.parse(e.extras),
            });
```

jaller94 (approved):
Ok, but please use SQL's LIMIT.

src/datastore/postgres/PgDatastore.ts
Half-Shot: totally forgot that was a thing. Clever!

Issue #501, opened 2020-09-23
Onboarding users with puppeting


Half-Shot:
Currently puppeting oneself requires knowing to message the bot user, and logging in. This however is not very user friendly. The ideal process would be:

- The bridge keeps a track of matrix users in it's rooms.
- If a user is in a room, or joins it and is neither puppeted nor has a DM with the bot already: it creates a DM
- The bot should then give a friendly (word-smithed) introduction to the bridge and explain that they need to authenticate
   to get the best experience.
- A link should be provided which will take them to a web page that gives them the option of signing into their Slack account, or signing up to the workspace if they haven't already.
- The oauth process continues as normal, and the user should get one final message once the process is complete.

Cadair:
> that they need to authenticate to get the best experience.

that they can authenticate if they want to be able to receive DMs / private channels from slack.

PR #506, opened 2020-09-24
Add message to onboard new users to the Slack bridge

Half-Shot:
This is to hopefully fix some of #501. It's not a complete solution but should hopefully make the on-boarding experience more pleasant.

image


templates/onboarding.html
```diff
@@ -0,0 +1,12 @@
...
+    👋 Hello there! You have just joined a room bridged to Slack, which means messages you send
+    will be automatically bridged to Slack, and messages sent from Slack will appear in the room.
+    This room serves as your control center for configuring the bridge, and is accessible
+    only to you.
+</p>
+
+<p>
+    The bridge is not yet connecting direct messages, private channels or various other nice
+    things like reactions to your account. In order to do that, please type <code>login</code> in order
+    to connect your account to the bridge. You may also type <code>help</code> for other commands.
```
jaller94, line 11: Technically typing doesn't do the job. You gotta press Enter as well. 🤓
```suggestion
    things like reactions to your account. In order to do that, please send me the message <code>login</code>
    in order to connect your account to the bridge. You may also send <code>help</code> for other commands.
```

templates/onboarding.txt
```diff
@@ -0,0 +1,9 @@
+Hello there! You have just joined a room bridged to Slack, which means messages you send
+will be automatically bridged to Slack, and messages sent from Slack will appear in the room.
+
+This room serves as your control center for configuring the bridge, and is visible
+only to you.
+
+The bridge is not yet connecting direct messages, private channels or various other nice
+things like reactions to your account. In order to do that, please type `login` in order
+to connect your account to the bridge. You may also type `help` for other commands.
```
jaller94, line 9:
```suggestion
The bridge is not yet connecting direct messages, private channels or various other nice
things like reactions to your account. In order to do that, please send the message `login`
in order to connect your account to the bridge. You may also send `help` for other commands.
```

src/Main.ts
```diff
@@ -64,6 +64,8 @@ export interface ISlackTeam {
 
 interface MetricsLabels { [labelName: string]: string; }
 
+type TimerFunc = (labels?: Partial<Record<string, string | number>> | undefined) => void;
```
jaller94, line 67: What does `Partial` give us here?
I think `Partial<Record<string, string | number>>` is the same as `Record<string, string | number>` as a parameter when you're not using any other property in the function.

src/Main.ts
```diff
@@ -362,7 +364,7 @@ export class Main {
         }
     }
 
-    public startTimer(name: string, labels: MetricsLabels = {}) {
+    public startTimer(name: string, labels: MetricsLabels = {}): TimerFunc {
```
jaller94, line 367: Nice! I should have introduced `TimerFunc`.

src/Main.ts
```diff
@@ -534,22 +508,84 @@ export class Main {
...
+                // duplicate at the end (reasoning: we only want one of the
+                // duplicated ev_id in the list, but we want it at the end)
+                recents[i] = recents[this.mostRecentEventIdIdx];
+                recents[this.mostRecentEventIdIdx] = ev.event_id;
+                log.warn("Ignoring duplicate ev: " + ev.event_id);
+                return;
+            }
+        }
+        this.mostRecentEventIdIdx = (this.mostRecentEventIdIdx + 1) % RECENT_EVENTID_SIZE;
+        recents[this.mostRecentEventIdIdx] = ev.event_id;
```
jaller94, line 582: This code is getting hard to understand.
May be time to make a helper class out of this cached list.
It's also hard to understand how this all behaves after a restart and if a busy bridge behaves worse because events are leaving the cache quicker. (I wish we would persist it.)

config/config.sample.yaml
```diff
@@ -126,6 +126,9 @@ provisioning:
 puppeting:
   # Should the bridge allow users to puppet their accounts
   enabled: false
+  # Should the bridge send users a onboarding message when
+  # they join a Slack channel for the first time.
+  onboard_users: false
```
jaller94, line 131: I think this should suggested to be on by default. 
```suggestion
  onboard_users: true
```

✨🐕‍🦺 First time user experience 💖🌟

jaller94 (approved):
What a great addition.

config/config.sample.yaml
Half-Shot: Okay, on for new bridges but off for existing?

src/Main.ts
Half-Shot: I copy pasted it from the library :shrug:

src/Main.ts
Half-Shot: :)

src/Main.ts
Half-Shot: Yeah, I'm wholly unconvinced by this code tbh. I've left it here for now because I'd rather not fight it, but we should use an LRU or something.

config/config.sample.yaml
jaller94: Yeah, lets default to off if this is undefined.

jaller94:
I'm currently merging this…

PR #508, opened 2020-09-24
Fix URL slashes again

Half-Shot:
A lesson learned that `path.resolve` will not resolve URLs for you.


src/OAuth2.ts
```diff
@@ -134,6 +135,6 @@ export class OAuth2 {
     }
 
     private makeRedirectURL(token: string): string {
-        return path.join(this.redirectPrefix, token, "authorize");
+        return url.resolve(this.redirectPrefix, `${token}/authorize`);
```
jaller94, line 138: Looks like `url.resolve` got deprecated and has "Stability: 0".
https://nodejs.org/api/url.html#url_url_resolve_from_to

As far as I see this is supposed to be the replacement.
https://developer.mozilla.org/en-US/docs/Web/API/URL_API
Didn't find what this translates to nor if this is supported in NodeJS, besides the documentation note to us this API instead.

jaller94 (changes requested):

src/OAuth2.ts
Half-Shot: ah fuck

Half-Shot: Now using URL, tests should ensure this is doing the right thing.

src/OAuth2.ts
```diff
@@ -134,6 +134,6 @@ export class OAuth2 {
     }
 
     private makeRedirectURL(token: string): string {
-        return path.join(this.redirectPrefix, token, "authorize");
+        return new URL(`${token}/authorize`, this.redirectPrefix).toString();
```
jaller94, line 137: The second parameter of `new URL()` seems to only allow the protocol and domain.
If you add a path (e.g. `https://redir_prefix/my-custom-oauth-endpoint`) the output becomes `https://redir_prefix/token/authorize`.

jaller94 (changes requested):
I think this introduces a regression where the path of `redir_prefix` gets cut off.

src/OAuth2.ts
Half-Shot: Oh, right you are. Well, failing that I guess we will have to use a library :|

jaller94 (approved):
Has a little merge conflict, but this solution to the issue like good to me. :+1:

Issue #488, opened 2020-09-14
👍️ and 👎️ reactions are double posted for puppeted Matrix users


jaller94:
**Describe the bug**
👍️ and 👎️ reactions are double posted for puppeted Matrix users.

**To Reproduce**
1. Be a Matrix user in a bridged room.
2. Enable puppeting (by writing "login" to the bridge bot and authenticating via the link it posts).
3. Write a message and react to it with either 👍️ or 👎️

**Expected behavior**
The reactions are visible once (as is the case with all other emojis), not twice (bridged back from Slack to Matrix).

**Screenshots**
![Screenshot_2020-09-14_14-42-13](https://user-images.githubusercontent.com/10872136/93087194-841a4f00-f698-11ea-951d-6175ffc8c801.png)

jaller94:
Current best bet (not 100% confirmed):
The unicode sequence that the Element client sends appears to be a different one than what `node-emoji`'s `.emojify()` returns.

jaller94:
In the following the output of `emoji.emojify()` is called "Slack". This is not what Slack sends (which would be ":thumbsup:") but what ":thumbsup:" gets converted to by the library we use.

** thumbs up **
![Screenshot_2020-09-24_21-16-05](https://user-images.githubusercontent.com/10872136/94189625-45ac3d80-feab-11ea-99ae-a85565240da1.png)

** thumbs down **
![Screenshot_2020-09-24_21-18-07](https://user-images.githubusercontent.com/10872136/94189797-81470780-feab-11ea-9b57-e034e1cbc713.png)

Page to convert these: https://onlineutf8tools.com/convert-bytes-to-utf8

jaller94:
Looks like Element appends 239 184 143 VARIATION SELECTOR-16 after the thumbs up and down.
https://www.utf8-chartable.de/unicode-utf8-table.pl

PR #509, opened 2020-09-24
Append /ufe0f to thumbsup and thumsdown reactions

jaller94:
Fixes: #488 

Related: https://github.com/vector-im/element-web/issues/13926

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -660,7 +660,11 @@ export class BridgedRoom {
             return;
         }
 
-        const reactionKey = emoji.emojify(`:${message.reaction}:`, getFallbackForMissingEmoji);
+        let reactionKey = emoji.emojify(`:${message.reaction}:`, getFallbackForMissingEmoji);
+        // Element uses the default thumbsup and thumbsdown reactions with an appended variant character.
+        if (reactionKey === '👍' || message.reaction === '👎') {
+            reactionKey += '\ufe0f'.normalize(); // VARIATION SELECTOR-16
+        }
 
         if (this.recentSlackMessages.includes(`reactadd:${reactionKey}:${message.user_id}:${message.item.ts}`)) {
             // We sent this, ignore.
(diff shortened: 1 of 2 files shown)


src/BridgedRoom.ts
```diff
@@ -660,7 +660,11 @@ export class BridgedRoom {
             return;
         }
 
-        const reactionKey = emoji.emojify(`:${message.reaction}:`, getFallbackForMissingEmoji);
+        let reactionKey = emoji.emojify(`:${message.reaction}:`, getFallbackForMissingEmoji);
+        // Element uses the default thumbsup and thumbsdown reactions with an appended variant character.
+        if (reactionKey === '👍' || message.reaction === '👎') {
```
Half-Shot, line 665:
```suggestion
        if (reactionKey === '👍' || reactionKey === '👎') {
```

Half-Shot: Spotted

Half-Shot (approved):
Otherwise excellent!

src/BridgedRoom.ts
```diff
@@ -660,7 +660,11 @@ export class BridgedRoom {
...
+            reactionKey += '\ufe0f'.normalize(); // VARIATION SELECTOR-16
```
Cadair, line 666: it's only the :+1: and :-1: emoji Element does this to?!

jaller94: I tested the other quick reactions and they work fine. (2 means that Matrix and Slack are sending the same character)
Screenshot_2020-09-24_21-56-39

Cadair: From the MSC: 

> When sending emoji reactions, the key field should include the colourful variation-16 when applicable.

why is it only applicable to append VS16 to :+1:  and :-1: ?!

tulir: It's not applicable to everything, but I'm 97% sure there's more than :+1:/:-1:

PR #515, opened 2020-09-25
Join slack client to channel if they aren't in it.

### LINKED ISSUE #512: Invite or autojoin a user if we get a not_in_channel
Particularly if the bridge is older and new joins were not synced to Slack, we should rectify this.

Half-Shot:
Fixes #512


jaller94 (approved):
Looks good to me.

PR #516, opened 2020-09-26
Simplify deduplication code

### LINKED ISSUE #514: recentMatrixEventIds code is needlessly complex
This code is getting hard to understand.
May be time to make a helper class out of this cached list.
It's also hard to understand how this all behaves after a restart and if a busy bridge behaves worse because events are leaving the cache quicker. (I wish we would persist it.)

_Originally posted by @jaller94 in https://github.com/matrix-org/matrix-appservice-slack/pull/506#r494926716_

Half-Shot:
Fixes #514 

I'm not entirely sure why we were deduplicating incoming events, because the bridge library and homeserver ensure that we never see the same event twice.

I've also added supressEcho, because I couldn't find any code where the bridge actually handles it's own events.


jaller94 (approved):
Looks good to me.

PR #521, opened 2020-09-29
Upgrade dependencies

jaller94:
I read through all changelogs of major updates:
* `@slack/logger` removed(?) a method `.getLogger()` which we don't use.
* `escape-string-regexp` made changes to how - (dash) is outputted, but we never care about the exact output.
* [`winston-daily-rotate-file` removed support for winston@2](https://github.com/winstonjs/winston-daily-rotate-file/releases/tag/v4.0.0) but we're using winston@3.
* [`yargs`@16](https://github.com/yargs/yargs/releases/tag/v16.0.0) changed its EMS API surface, resorted single character aliases to show first in the help, and removed a rebase helper. I tested that the `help` and `list` command still work as expected.


Half-Shot (approved):
Slightly worried about yargs, but if it still works then LGTM.

PR #522, opened 2020-09-29
WIP: End all emoji reactions on VS16

jaller94:
This does not seem to be compatible with Element at the moment. Element does end some but not all emojis with VS16.

Also, as of today [MSC2677](https://github.com/matrix-org/matrix-doc/pull/2677) hasn't been merged.

### DIFF
--- a/changelog.d/522.bugfix
+++ b/changelog.d/522.bugfix
@@ -1,0 +1,1 @@
+End all emoji reactions on VS16 in compliance with MSC2677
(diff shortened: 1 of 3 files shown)


Half-Shot:
The MSC states:

>When sending emoji reactions, the key field should include the colourful variation-16 when applicable.

which seems to line up with:

>This does not seem to be compatible with Element at the moment. Element does end some but not all emojis with VS16.

?

Cadair:
Please define "when applicable" :stuck_out_tongue:

jaller94:
I thought that "when applicable" means emoji characters vs. text characters.
However, it's emoji characters with a text variant vs. those without.

> you wouldn't put VS16 on all emojis - only when there's a text variant
> ~Dave

I will change the PR accordingly.

jaller94:
I don't think this needs work unless we discover a tangible bug.

PR #523, opened 2020-09-29
WIP: Refactor BridgedRoom into two classes

Half-Shot:
We have a lot of code that tries to handle both webhooks and richer bridged channels, however webhooks can only really be used to deliver simple messages. In order to make future work easier on the richer channels, the code will be split into two classes.


Half-Shot:
This has too many conflicts now. Still something we want to do, but might be a future change.

PR #525, opened 2020-10-01
Do not fail to bridge slack message if an alias doesn't exist

### LINKED ISSUE #524: Slack message can fail to bridge if transforming a channel mention fails
The bridge will try to fetch `m.room.canonical_alias` state for the room mentioned (in order to get the right alias), and if it fails to get the event then the code will throw and subsequently fail to bridge the message. This is obviously not ideal, and we should fall back to textual representation.

Half-Shot:
Fixes #524

### DIFF
--- a/src/BaseSlackHandler.ts
+++ b/src/BaseSlackHandler.ts
@@ -171,11 +171,14 @@ export abstract class BaseSlackHandler {
             // If we bridge the room, attempt to look up its canonical alias.
             if (room !== undefined) {
                 const client = this.main.botIntent.getClient();
-                const canonical = await client.getStateEvent(room.MatrixRoomId, "m.room.canonical_alias");
-                if (canonical !== undefined && canonical.alias !== undefined) {
-                    text = text.slice(0, match.index) + canonical.alias + text.slice(match.index + match[0].length);
-                } else {
+                try {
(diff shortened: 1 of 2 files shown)


src/BaseSlackHandler.ts
```diff
@@ -171,11 +171,14 @@ export abstract class BaseSlackHandler {
...
-                    text = text.slice(0, match.index) + canonical.alias + text.slice(match.index + match[0].length);
-                } else {
+                try {
+                    const canonical = await client.getStateEvent(room.MatrixRoomId, "m.room.canonical_alias");
+                    if (canonical !== undefined && canonical.alias !== undefined) {
+                        text = text.slice(0, match.index) + canonical.alias + text.slice(match.index + match[0].length);
+                    }
+                } catch (ex) {
                     // If we can't find a canonical alias fall back to just the Slack channel name.
+                    log.debug(`Room ${room.MatrixRoomId} does not have a canonical alias set.`);
```
jaller94, line 181: Lets log ex here. The error may not come from the promise failing as there's more in the try block.

"have an alias set" is ambiguous as "set" can be a verb or a noun.

```suggestion
                    log.debug(`Room ${room.MatrixRoomId} does not have a canonical alias.`, ex);
```

jaller94 (approved):
Looks good to me, with one exception.

PR #528, opened 2020-10-01
Hide typing from puppeted users

Half-Shot:
If a puppeted user starts typing on Matrix, the bridge will currently reflect that as their remote user typing. We should hide the typing notifications for users we puppet.


jaller94 (approved):
Looks good to me.

PR #531, opened 2020-10-03
Minimize docker build time

erdnaxeli:
We copy only npm related files first, so we can use the cache and not run again
`npm install` when only the code changes.

Signed-off-by: Alexandre Morignot <erdnaxeli@cervoi.se>

This speeds up a bit development time, when using the docker image to test (as I do).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,25 @@
 FROM node:14-alpine AS BUILD
-COPY . /src
 
 # git is needed to install Half-Shot/slackdown
 RUN apk add git
 WORKDIR /src
+
+COPY package.json package-lock.json /src/
 RUN npm install
+COPY . /src
 RUN npm run build
 
 FROM node:14-alpine
 
 VOLUME /data/ /config/
 
-COPY package.json /usr/src/app/
-COPY package-lock.json /usr/src/app/
+WORKDIR /usr/src/app
+COPY package.json package-lock.json /usr/src/app/
+RUN apk add git && npm install --only=production
 
 COPY --from=BUILD /src/config /usr/src/app/config
 COPY --from=BUILD /src/templates /usr/src/app/templates
 COPY --from=BUILD /src/lib /usr/src/app/lib
-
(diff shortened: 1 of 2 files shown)


changelog.d/531.feat
```diff
@@ -0,0 +1 @@
+Allow docker to reuse cache when building the image by copying only npm related file before running `npm install`
```
Half-Shot, line 1: s/.feat/.feature/

Dockerfile
```diff
@@ -1,27 +1,26 @@
 FROM node:14-alpine AS BUILD
-COPY . /src
 
 # git is needed to install Half-Shot/slackdown
 RUN apk add git
 WORKDIR /src
+
+COPY package.json package-lock.json /src/
 RUN npm install
+COPY . /src
```
Half-Shot, line 9: If we do this anyway, what's the advantage of two copy operations?

Dockerfile
```diff
@@ -1,27 +1,26 @@
...
 
 FROM node:14-alpine
 
 VOLUME /data/ /config/
 
-COPY package.json /usr/src/app/
-COPY package-lock.json /usr/src/app/
+WORKDIR /usr/src/app
+COPY package.json package-lock.json /usr/src/app/
+RUN apk add git && npm install --only=production
```
Half-Shot, line 18: I'm not sure how this makes things faster, could you explain how the ordering affects things?

Dockerfile
erdnaxeli: The advantage is that I first only copy the package.json and package-lock.json files. So if they didn't change since the last build docker can use the cache for this step, hence it can also use the cache for the `RUN npm install` step.

Dockerfile
erdnaxeli: Same idea as before, we run `npm install` before copying the whole code, so if the code change but pacakge.json and package-lock.json don't docker can use the cache and we don't to have to run npm install again.

Half-Shot (approved):

PR #544, opened 2020-11-18
Fix public rooms being created without aliases

Half-Shot:
This was using the wrong option as pointed out by @MadLittleMods (https://gitlab.com/gitterHQ/webapp/-/merge_requests/2053).

We need to be careful here, because I can imagine making this work would introduce a breaking change insofar that if the aliases start working, the HS may reject the request because the registration file hasn't included the correct namespace.

The configuration to set aliases is behind a configuration flag, and I've ensured that generated registration files will include the correct alias prefixes.

### DIFF
--- a/src/TeamSyncer.ts
+++ b/src/TeamSyncer.ts
@@ -491,7 +491,7 @@ export class TeamSyncer {
                 name: `#${channel.name}`,
                 topic,
                 visibility: isPublic ? "public" : "private",
-                room_alias: alias,
+                room_alias_name: alias,
                 preset: isPublic ? "public_chat" : "private_chat",
                 invite: inviteList,
                 initial_state: extraContent,
--- a/src/app.ts
+++ b/src/app.ts
@@ -39,6 +39,12 @@ const cli = new Cli({
         reg.setAppServiceToken(AppServiceRegistration.generateToken());
         reg.setSenderLocalpart("slackbot");
         reg.addRegexPattern("users", `@${config.username_prefix}.*:${config.homeserver.server_name}`, true);
+        const teamSyncEntries = config.team_sync && Object.values(config.team_sync) || [];
+        for (const teamEntry of teamSyncEntries) {
+            if (teamEntry.channels?.alias_prefix) {
+                reg.addRegexPattern("aliases", `#${teamEntry.channels?.alias_prefix}.*:${config.homeserver.server_name}`, true);
+            }
+        }
         callback(reg);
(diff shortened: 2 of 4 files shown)


src/TeamSyncer.ts
```diff
@@ -470,7 +470,7 @@ export class TeamSyncer {
                 name: `#${channel.name}`,
                 topic,
                 visibility: isPublic ? "public" : "private",
-                room_alias: alias,
+                room_alias_name: alias,
```
jaller94, line 473: I don't know how to verify this option name other than confirming that it is also used in tests of the bridge lib.
https://github.com/matrix-org/matrix-appservice-bridge/blob/ac522565ecc71d43d5cec7f972372acf92877fdd/spec/integ/bridge.spec.js#L216

src/app.ts
```diff
@@ -36,6 +36,12 @@ const cli = new Cli({
         reg.setAppServiceToken(AppServiceRegistration.generateToken());
         reg.setSenderLocalpart("slackbot");
         reg.addRegexPattern("users", `@${config.username_prefix}.*:${config.homeserver.server_name}`, true);
+        const teamSyncEntries = config.team_sync && Object.values(config.team_sync) || [];
+        for (const teamEntry of teamSyncEntries) {
+            if (teamEntry.channels?.alias_prefix) {
+                reg.addRegexPattern("aliases", `#${teamEntry.channels?.alias_prefix}.*:${config.homeserver.server_name}`, true);
+            }
+        }
```
jaller94, line 44: This allows the app service to control all aliases matching that regex?. Good to know.

You mention this being a breaking change. That's because as long as admins don't re-generate the registration file, the appservice doesn't have permission to create/modify any aliases, right?

jaller94 (approved):

jaller94:
The tests are failing because it's not merged with the latest `develop`. It's unrelated to this PR.

src/TeamSyncer.ts
Half-Shot: Sort of thing we should really have a type for. Wish the js-sdk exported types that we could re-use.

MadLittleMods: For reference and extra confidence, we also use this option in the Gitter bridge, https://gitlab.com/gitterHQ/webapp/-/blob/0b9f181394767c4cff7c131f6da9ec30314daa0d/modules/matrix-bridge/lib/matrix-utils.js#L84

PR #565, opened 2021-01-15
Add supporting code for latest enc bridge efforts

Half-Shot:
Depends on https://github.com/matrix-org/matrix-appservice-bridge/pull/285

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "chai": "^4.2.0",
     "escape-string-regexp": "^4.0.0",
     "matrix-appservice": "^0.7.1",
-    "matrix-appservice-bridge": "^2.4.1",
+    "matrix-appservice-bridge": "github:matrix-org/matrix-appservice-bridge#729a604bd4d3dcddf1b901907a7bd9e1f2caa625",
     "minimist": "^1.2.5",
     "nedb": "^1.8.0",
     "node-emoji": "^1.10.0",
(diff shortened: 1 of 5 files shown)


src/datastore/postgres/PgDatastore.ts
```diff
@@ -497,6 +498,10 @@ export class PgDatastore implements Datastore, ClientEncryptionStore {
         await this.postgresDb.none(statement, params);
```
jaller94, line 498: The syncToken cannot be set here. Is that intentional?

jaller94 (approved):
I checked the type ClientEncryptionSession on in the -bridge repo on that referenced commit.

package.json
```diff
@@ -36,7 +36,7 @@
     "chai": "^4.2.0",
     "escape-string-regexp": "^4.0.0",
     "matrix-appservice": "^0.7.1",
-    "matrix-appservice-bridge": "^2.4.1",
+    "matrix-appservice-bridge": "github:matrix-org/matrix-appservice-bridge#729a604bd4d3dcddf1b901907a7bd9e1f2caa625",
```
jaller94, line 39: The CI doesn't seem happy about this.
```
error TS2307: Cannot find module 'matrix-appservice-bridge' or its corresponding type declarations.
```

jaller94 (approved):
Looks good to me.

PR #566, opened 2021-02-07
fix wrong config key name

phpwutz:
according to the code and the spec it's supposed to be log_level instead of logging.
really annoying when you try to debug why RTM doesn't work and the setting is buggy itself :P


Half-Shot:
Looks good, can you add a changelog and sign off as per https://github.com/matrix-org/matrix-appservice-slack/blob/develop/CONTRIBUTING.md#making-pull-requests?

Issue #569, opened 2021-02-16
Add endpoint for room info


jaller94:
**Is your feature request related to a problem? Please describe.**
As a bridge admin I want to ask the bridge how many Slack users are in a specific room.

**Describe the solution you'd like**
An API endpoint similar to `/metrics` that gives me room info.

I give it:
 * Slack workspace ID
 * Slack channel ID

It provides me:
 * number of Slack users in that room

**Describe alternatives you've considered**
* Having a Prometheus gauge in `/metrics`. But that would query too much unused data when a bridge has a ton of rooms.

**Additional context**
![Untitled(2)](https://user-images.githubusercontent.com/10872136/108079278-f25fa500-706e-11eb-8a6e-264133127021.png)

jaller94:
No longer relevant and we have other (Prometheus-based) metrics now.

PR #571, opened 2021-02-22
Add provisioning endpoint to get Slack channel info

jaller94:
A first idea on how to solve #569

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1212,8 +1212,7 @@ export class Main {
     //  * Associate a webhook_uri to an existing instance
     public async actionLink(opts: {
         matrix_room_id: string,
-        slack_webhook_uri?: string,
-        slack_channel_id?: string,
+        slack_channel_id: string,
         slack_bot_token?: string,
         team_id?: string,
     }): Promise<BridgedRoom> {
@@ -1228,12 +1227,6 @@ export class Main {
 
         if (existingChannel) {
             throw Error("Channel is already bridged! Unbridge the channel first.");
-        }
-
-        if (!opts.team_id && !opts.slack_bot_token) {
-            if (!opts.slack_webhook_uri) {
-                throw Error("Neither a team_id nor a slack_bot_token were provided");
-            }
         }
 
         if (opts.slack_bot_token) {
@@ -1310,16 +1303,12 @@ export class Main {
             room = existingRoom;
         }
 
-        if (opts.slack_webhook_uri) {
-            room.SlackWebhookUri = opts.slack_webhook_uri;
-        }
-
         if (opts.slack_channel_id) {
             room.SlackChannelId = opts.slack_channel_id;
         }
 
-        if (!room.SlackChannelId && !room.SlackWebhookUri) {
-            throw Error("Missing webhook_id OR channel_id");
+        if (!room.SlackChannelId) {
+            throw Error("Missing channel_id");
         }
 
         if (slackClient) {
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1131,6 +1131,81 @@ export class Main {
         }
     }
 
+    public async getChannelInfo(opts: {
+        slack_webhook_uri?: string,
+        slack_channel_id: string,
+        slack_bot_token?: string,
+        team_id?: string,
+    }): Promise<ConversationsInfoResponse|void> {
+        let slackClient: WebClient | undefined;
+        let teamEntry: TeamEntry | null = null;
+        let teamId: string = opts.team_id!;
+
+        const existingChannel = opts.slack_channel_id ? this.rooms.getBySlackChannelId(opts.slack_channel_id) : null;
+
+        if (existingChannel) {
+            throw Error("Channel is already bridged! Unbridge the channel first.");
+        }
+
+        if (!opts.team_id && !opts.slack_bot_token) {
+            if (!opts.slack_webhook_uri) {
+                throw Error("Neither a team_id nor a slack_bot_token were provided");
+            }
+        }
+
+        if (opts.slack_bot_token) {
+            if (!opts.slack_bot_token.startsWith("xoxb-")) {
+                throw Error("Provided token is not a bot token. Ensure the token starts with xoxb-");
+            }
+            // We may have this team already and want to update the token, or this might be new.
+            // But first check that the token works.
+            try {
+                teamId = await this.clientFactory.upsertTeamByToken(opts.slack_bot_token);
+                log.info(`Found ${teamId} for token`);
+            } catch (ex) {
+                log.error("Failed to action link because the token couldn't used:", ex);
+                throw Error("Token did not work, unable to get team");
+            }
+        }
+
+        // else, assume we have a teamId
+        if (teamId) {
+            try {
+                slackClient = await this.clientFactory.getTeamClient(teamId);
+            } catch (ex) {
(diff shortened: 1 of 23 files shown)


jaller94:
Judging by the CI, no. I'll take this on tomorrow.

jaller94:
@Half-Shot This is ready for review.

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
         }
     }
 
+    public async getChannelInfo(opts: {
+        slack_channel_id: string,
+        slack_bot_token: string,
+        team_id: string,
+    }): Promise<ConversationsInfoResponse|void> {
+        let slackClient: WebClient | undefined;
```
Half-Shot, line 1128: Nit but, the spacing here is a bit odd.

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+        let teamEntry: TeamEntry | null = null;
+        let teamId: string = opts.team_id;
+
+        const existingChannel = opts.slack_channel_id ? this.rooms.getBySlackChannelId(opts.slack_channel_id) : null;
+
+        if (existingChannel) {
+            throw Error("Channel is already bridged! Unbridge the channel first.");
+        }
+
+        if (opts.slack_bot_token) {
```
Half-Shot, line 1138: We shouldn't be passing the slack token in here

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+                teamId = await this.clientFactory.upsertTeamByToken(opts.slack_bot_token);
+                log.info(`Found ${teamId} for token`);
+            } catch (ex) {
+                log.error("Failed to action link because the token couldn't used:", ex);
+                throw Error("Token did not work, unable to get team");
+            }
+        }
+
+        // else, assume we have a teamId
+        if (teamId) {
```
Half-Shot, line 1154: We should always have a teamId. If the team doesn't exist yet, the calling function should make sure we do create it.

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+
+            teamEntry = await this.datastore.getTeam(teamId);
+            if (!teamEntry) {
+                throw Error("Team ID provided, but no team found in database");
+            }
+        }
+
+        let channelInfo: ConversationsInfoResponse | undefined;
+        if (slackClient && opts.slack_channel_id && opts.team_id) {
+            // PSA: Bots cannot join channels, they have a limited set of APIs https://api.slack.com/methods/bots.info
```
Half-Shot, line 1170: This comment is redundant. We aren't trying to join a channel for channel info

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
         }
     }
 
+    public async getChannelInfo(opts: {
```
Half-Shot, line 1123: I'd drop the opts object here and have them as real parameters, it's a bit of an ugly design pattern.

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+            }
+
+            teamEntry = await this.datastore.getTeam(teamId);
+            if (!teamEntry) {
+                throw Error("Team ID provided, but no team found in database");
+            }
+        }
+
+        let channelInfo: ConversationsInfoResponse | undefined;
+        if (slackClient && opts.slack_channel_id && opts.team_id) {
```
Half-Shot, line 1169: As per above, we can assume a teamId and a channelId

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+            if (!channelInfo.ok) {
+                if (channelInfo.error === 'channel_not_found') {
+                    return;
+                }
+                log.error(`conversations.info for ${opts.slack_channel_id} errored:`, channelInfo.error);
+                throw Error("Failed to get channel info");
+            }
+        }
+
+        if (opts.slack_channel_id &&
```
Half-Shot, line 1182: We can assume a channelId at this point

src/Main.ts
```diff
@@ -1120,6 +1120,74 @@ export class Main {
...
+            this.allowDenyList.allowSlackChannel(opts.slack_channel_id, channelInfo?.channel.name) !== DenyReason.ALLOWED) {
+            log.warn(`Channel ${opts.slack_channel_id} is not allowed to be bridged`);
```
Half-Shot, line 1184: Probably not a warning. Also not sure the error is helpful as we're not trying to bridge it.

src/Provisioning.ts
```diff
@@ -282,6 +282,57 @@ export class Provisioner {
         });
     }
 
+    @command("user_id", "channel_id", "team_id", "bot_token")
```
Half-Shot, line 285: We don't need the bot_token

src/Provisioning.ts
```diff
@@ -282,6 +282,57 @@ export class Provisioner {
...
+        if (!(await this.main.matrixUserInSlackTeam(opts.team_id, userId))) {
+            return Promise.reject({
+                code: HTTP_CODES.FORBIDDEN,
+                text: `${userId} is not in this team.`,
+            });
+        }
+
+        let channelInfo;
+        try {
+            channelInfo = await this.main.getChannelInfo(opts);
```
Half-Shot, line 314: You could simplyify this code a lot by having the res.json part inside the try catch statement, 
```suggestion
            const channelInfo = await this.main.getChannelInfo(opts);
            if (!channelInfo) {
                return res.status(HTTP_CODES.NOT_FOUND).json({
                    message: 'Slack channel not found',
                });
            }
            res.json({
                name: channelInfo.channel.name,
                memberCount: channelInfo.channel.num_members,
            });
```

Half-Shot (commented):
Looks plausible but there is a lot of cruft to remove here.

src/Main.ts
jaller94: Removed everything related to botToken for getChannelInfo.

src/Main.ts
jaller94: Removed the spaces

src/Main.ts
jaller94: Unwrapped the parameters.

src/Main.ts
jaller94: The calling function now ensures it's a non-empty string.
Removed the condition, keeping its content.

src/Main.ts
jaller94: Removed the two conditions.

src/Main.ts
jaller94: Removed the comment.

src/Main.ts
jaller94: Removed the condition.

src/Main.ts
jaller94: Replaced with an empty return value.
The API now says:
`{
    message: 'Slack channel not found or not allowed to be bridged',
}`

src/Provisioning.ts
jaller94: Removed

src/Provisioning.ts
jaller94: Applied suggestion.

src/Main.ts
jaller94: Suggestion to differentiate between "not found" and "not allowed" because I don't think this is confidential.

Half-Shot (approved):

PR #573, opened 2021-03-01
docs: tweaks install instructions re: port

vincentlaucy:
Hi first of all thanks for this great life-saver project!

I installed this to my server recently, some feedback regarding the docs

my setup env:
- terraform to provision homeserver/this bridge
- docker-compose on ec2
- db: rds

## feedback re: port
- I spent sometime to understand the port to be used. It is more obvious when I read the log of runnable process and check the source code. According to 
https://github.com/matrix-org/matrix-appservice-slack/issues/436 the default port is 5858, shall we provide in the instruction as well? (as changed in this PR)
- I understand it is the port homeserver will connect to, while the name MATRIX_PORT sounds a port homeserver is listening to. Potentially HOMESERVER_PORT SYNAPSE_PORT are more specific?

- By "specify the direct link", and generating the config with docker, it might not be too obvious that if refers to the "url field" in the config yml when I decided to modify directly. (added a note in this PR)

- changed the file name as `slack-registration.yaml` is the recommended name, not `slack.yaml`
 
### /link
- Probably there are other issues pointing to this already. I was confused link to be a `/link` command and struggled when it didnt show up. There are nice feedback on the correctness of commands once the appservice is installed, so we can probably hint user to look for this to hint whether it is problem of appservice process or problem of access configurations. 

## some thoughts for terraform/docker-compose
- I found out https://github.com/spantaleev/matrix-docker-ansible-deploy only after I have done the deployment. Shall we add a link to that repo? 

- Will a sample terraform/ docker-compose sample of interest? There are some networking tweaks e.g. listen to `"http://appservice:5858"`  instead of localhost, tweaking the volumes mount point etc

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -51,7 +51,7 @@ ever stuck, you can post a question in the
 1. Create a new Matrix room to act as the administration control room. Note its
    internal room ID (Example: !abcdefg12345hijk:coolserver.com).
 
-1. Decide on a spare local TCP port number to use. It will listen for messages
+1. Decide on a spare local TCP port number to use. The Default is 5858.  It will listen for messages
    from Matrix and needs to be visible to the homeserver. Take care to configure
    firewalls appropriately. This port will be notated as `$MATRIX_PORT` in
    the remaining instructions.
@@ -66,9 +66,11 @@ ever stuck, you can post a question in the
   
   1. For `matrix_admin_room`, enter the internal room ID of the administration control
      room (Example: !abcdefg12345hijk:coolserver.com).
+  
+  1. For `homeserver.appservice_port`, specify the above $MATRIX_PORT if you would like to override
 
 1. Generate the appservice registration file. This will be used by the
-   Matrix homeserver. Here, you must specify the direct link the
+   Matrix homeserver. Here, you must specify the direct link (the field url) the
    **Matrix Homserver** can use to access the bridge, including the Matrix
    port it will send messages through (if this bridge runs on the same
    machine you can use `localhost` as the `$HOST` name):
@@ -78,7 +80,7 @@ ever stuck, you can post a question in the
    
 ```sh
 $ docker run -v /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \ 
-    -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack.yaml
+    -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack-registration.yaml
 ```
 
 1. Start the actual application service:


docs/getting_started.md
```diff
@@ -51,7 +51,7 @@ ever stuck, you can post a question in the
 1. Create a new Matrix room to act as the administration control room. Note its
    internal room ID (Example: !abcdefg12345hijk:coolserver.com).
 
-1. Decide on a spare local TCP port number to use. It will listen for messages
+1. Decide on a spare local TCP port number to use. The Default is 5858.  It will listen for messages
```
Half-Shot, line 54:
```suggestion
1. Decide on a spare local TCP port number to use. The default is 5858.  It will listen for messages
```

docs/getting_started.md
```diff
@@ -66,9 +66,11 @@ ever stuck, you can post a question in the
   
   1. For `matrix_admin_room`, enter the internal room ID of the administration control
      room (Example: !abcdefg12345hijk:coolserver.com).
+  
+  1. For `homeserver.appservice_port`, specify the above $MATRIX_PORT if you would like to override
 
 1. Generate the appservice registration file. This will be used by the
-   Matrix homeserver. Here, you must specify the direct link the
+   Matrix homeserver. Here, you must specify the direct link (the field url) the
```
Half-Shot, line 73:
```suggestion
   Matrix homeserver. Here, you must specify the direct link (the url field) the
```

Half-Shot (approved):
This looks fine, be sure to add a changelog entry here

Half-Shot:
ping @vincentlaucy

docs/getting_started.md
```diff
@@ -66,9 +66,11 @@ ever stuck, you can post a question in the
   
   1. For `matrix_admin_room`, enter the internal room ID of the administration control
      room (Example: !abcdefg12345hijk:coolserver.com).
+  
+  1. For `homeserver.appservice_port`, specify the above $MATRIX_PORT if you would like to override
```
jaller94, line 70:
```suggestion
  1. For `homeserver.appservice_port`, specify the above $MATRIX_PORT if you overwrote the default.
```

jaller94:
@Half-Shot Do we need to insist on the Sign-Off?

@vincentlaucy To ensure all code in this project has the same license, please modify the following line and add it to the PR description:
`Signed-off-by: Your Name <your@email.example.org>`
More info: https://github.com/matrix-org/synapse/blob/v1.4.0/CONTRIBUTING.rst#sign-off

Half-Shot:
> @Half-Shot Do we need to insist on the Sign-Off?
> 
> @vincentlaucy To ensure all code in this project has the same license, please modify the following line and add it to the PR description: `Signed-off-by: Your Name <your@email.example.org>` More info: https://github.com/matrix-org/synapse/blob/v1.4.0/CONTRIBUTING.rst#sign-off

Yep, we do!

jaller94 (changes requested):
Sign-Off is missing.

vincentlaucy:
Thanks appended the issue with sign off

zhubonan:
Hi it would be great to have this merged. I have been following the current docs and using docker to setup but it did not work out right, mainly due to the port configuration and the default name of the registration file. 

I managed to solve the issues eventually but then I found the changes I needed are all in this PR. 😂  

The other thing I found is that to use a custom port (other than 5858) the docker has to run with the `-p` option explicitly, just like the system installed approach.

PR #575, opened 2021-03-05
Update matrix-appservice-bridge-2.6.0

Half-Shot:


src/tests/integration/HttpTests.ts
```diff
@@ -30,8 +30,8 @@ describe("HttpTests", () => {
     it("will respond 201 to a health check", async () => {
```
jaller94, line 30:
```suggestion
    it("will respond 200 to a health check", async () => {
```

jaller94 (approved):
/health became an endpoint of matrix-appservice-bridge.? ok

Technically the PR should be called "Update matrix-appservice-bridge-2.6.0-rc1".

PR #576, opened 2021-03-10
Only encrypt private rooms

Half-Shot:
We should never have encrypted public rooms

### DIFF
--- a/changelog.d/576.bugfix
+++ b/changelog.d/576.bugfix
@@ -1,0 +1,1 @@
+Autocreated public rooms will no longer be encrypted by default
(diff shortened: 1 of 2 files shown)


jaller94 (approved):
Sounds like a good short-term mitigation of errors.
However, I think that there are valid use cases for encryption of public rooms, e.g. room is "public" but restricted to a closed homeserver.

Half-Shot:
> However, I think that there are valid use cases for encryption of public rooms, e.g. room is "public" but restricted to a closed homeserver.

This is the automatic encryption system. You can of course manually encrypt rooms yourself by toggling the bit.

PR #577, opened 2021-03-10
Allow public channel syncing to be disabled independently of private

Half-Shot:
This is to allow people to sync their private channels without having to worry about their public ones also being bridged (for instance if they run a seperate sync tool for public channels aka @Cadair )


Cadair (approved):
Looks sensible to me

PR #579, opened 2021-03-16
ProvisioningAPI.channelinfo: Differentiate between not found and not allowed

jaller94:
I don't think we consider this to be confidential for Matrix users who have a puppet on the Slack server.

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1123,9 +1123,9 @@ export class Main {
     public async getChannelInfo(
         slackChannelId: string,
         teamId: string,
-    ): Promise<ConversationsInfoResponse|void> {
+    ): Promise<ConversationsInfoResponse|'channel_not_allowed'|'channel_not_found'> {
         let slackClient: WebClient|undefined;
-        let teamEntry: TeamEntry | null = null;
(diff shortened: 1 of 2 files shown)


src/Main.ts
```diff
@@ -1123,9 +1123,9 @@ export class Main {
     public async getChannelInfo(
         slackChannelId: string,
         teamId: string,
-    ): Promise<ConversationsInfoResponse|void> {
-        let slackClient: WebClient|undefined;
-        let teamEntry: TeamEntry | null = null;
+    ): Promise<ConversationsInfoResponse|'channel_not_allowed'|'channel_not_found'> {
```
Half-Shot, line 1126: Hmm, if we had to do to it this way I'd rather a `{error: string}` object response, just for extensibility reasons.

Half-Shot (approved):
Yeah, I think this is fine. Not allowed isn't terribly confidential :)

src/Main.ts
jaller94: Because this is an internal function, used only once, and the bridge isn't a library, I think a string is fine.
I considered an enum, but even that seemed overkill for just two error states with no additional info.

PR #582, opened 2021-04-07
Move tests to /tests

Half-Shot:
This PR moves tests to the `/tests` directory and avoids building them when building the project. We do this because:
- It's arguably cleaner to have the tests outside the source directory (and we do this for all other projects in the matrix-org ecosystem).
- It avoids us having to compile the tests when doing a build.
- There is no need to recompile the code when doing a test run, because ts-node will do that implicitly.


tsconfig.json
```diff
@@ -14,5 +14,9 @@
   },
   "include": [
     "src/**/*"
+  ],
+  "exclude": [
+    "tests/**/*",
```
jaller94, line 19: Is this needed if only `src` is included?
I guess someone would have to explicitly import from the tests folder.

jaller94 (approved):
I like to have unit tests right next to the source file for ease during development.
However, I'm not interested in bike-shedding over this and like the benefits you pointed out.

Looks good from a code perspective.

tsconfig.json
Half-Shot: It's actually not required :partying_face:  but removing *include* would cause it to attempt to build tests strangely.

PR #583, opened 2021-04-07
Validate sample and full config

Half-Shot:
Depends on #582 

To try and avoid having issues where we ship a release with a broken sample config


jaller94 (approved):
Nice to have some tests and also a regular and complete sample config. :+1:

PR #587, opened 2021-04-09
Config samples: Corrected rtm.logging to rtm.log_level; Schema: rtm.log_level: "off" -> "silent"

jaller94:
Replaces #566 by fixing both config samples


jaller94:
The schema wants it to be "off" instead of "silence" but the code only knows "silent".

Half-Shot (approved):

PR #601, opened 2021-05-17
Make the team sync hint to channel admins configurable

JanLuca:
Since the hint how to invite the bridge could be not wanted there should
be a option to disable this. This commit introduce a simple boalean flag
which makes the message configuarable for users.

Signed-off-by: Jan Luca Naumann <j.naumann@fu-berlin.de>


Half-Shot (approved):
No concerns from me, this looks great.

PR #605, opened 2021-07-20
Update packages (July edition)

Half-Shot:
Notably:

- Migrated to @alloc/quick-lru as we don't want to use ESM, which quick-lru@6.0.0 uses.
- We've held back `escape-string-regexp` as 5.0.0 also depends on ESM, and there isn't an alternative package.
- aaaand the same for p-queue. grrr


src/AdminCommands.ts
```diff
@@ -372,12 +372,15 @@ export class AdminCommands {
...
                 let matched = false;
-                this.yargs.parse(argv, {
+                const possiblePromise = this.yargs.parse(argv, {
                     // eslint-disable-next-line @typescript-eslint/no-unused-expressions
                     completed: (err) => { err ? reject(err) : resolve(true); },
                     matched: () => { matched = true; },
                     respond,
                 }, (err) => {  if (err !== null) { reject(err); } });
+                if (possiblePromise instanceof Promise) {
+                    possiblePromise.catch((ex) => reject(ex));
```
tadzik, line 382: Give than we synchronously `resolve(false)` below this, I don't think we can rely on this actually catching the error?

Maybe a try-await would fit better here, just to add to a stew of having every single async mechanism in this method :D It is a bit nuts. Is this fixing some bug we've discovered?

tadzik (changes requested):

src/AdminCommands.ts
Half-Shot: I think we've improved this immensely now

tadzik: We have indeed!

tadzik (approved):

PR #606, opened 2021-07-20
Fix crash when a join/leave on startup fails

Half-Shot:
Depends on #605 (just to ensure we use the right ver of p-queue)

If the promise rejects inside the queue, it bubbles up and causes the Node process to exit because it was not caught. So this PR now catches these.


src/SlackGhost.ts
```diff
@@ -62,6 +62,9 @@ export class SlackGhost {
         private datastore: Datastore,
         public readonly slackId: string,
         public readonly teamId: string|undefined,
+        /**
+         * The Matrix userId
```
tadzik, line 66: Maybe we ought to just rename it if it needs this comment?

src/TeamSyncer.ts
```diff
@@ -392,17 +393,29 @@ export class TeamSyncer {
...
+            try {
+                await ghost.intent.join(roomId);
+            } catch (ex) {
+                log.warn(`Failed to join ${ghost.userId} to ${roomId}`);
+            }
+        }));
 
         // Leave users who are joined
-        leftUsers.forEach(async(userId) => queue.add(async() => this.main.getIntent(userId).leave(roomId)));
+        void queue.addAll(leftUsers.map((ghost) => async () => {
```
tadzik, line 412: If we're sure that this will never fail and handling rejections is unnecessary, it should still be safer to `catch()` and fatal() what we get, in case the behaviour of `join()` or `leave()` changes in the future and causes this to silently eat errors.

tadzik (changes requested):

tadzik (approved):
Looks great \o/

PR #612, opened 2021-09-03
Implement RMAU limits and bridge blocking

tadzik:
Sister PR with https://github.com/matrix-org/matrix-appservice-irc/pull/1472, based on https://github.com/matrix-org/matrix-appservice-bridge/pull/350.

Drafting since I'm planning to move some code duplicated between -slack and -irc into bridge to reduce duplication.

### DIFF
--- a/config/config.sample-complete.yaml
+++ b/config/config.sample-complete.yaml
@@ -8,6 +8,9 @@ homeserver:
 
 username_prefix: "slack_"
 
+
+RMAU_limit: 100
+
 db:
   engine: "postgres"
   connectionString: "postgresql://slackbridge_user:somethingverysecret@localhost/slack_bridge"
--- a/config/config.sample-complete.yaml
+++ b/config/config.sample-complete.yaml
@@ -9,7 +9,7 @@ homeserver:
 username_prefix: "slack_"
 
 
-RMAU_limit: 100
+rmau_limit: 100
 
 db:
   engine: "postgres"
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -216,7 +216,9 @@ export class Main {
             controller: {
                 onEvent: async(request) => {
                     const ev = request.getData();
-                    if (this.bridgeBlocker?.isBlocked) {
+                    const isAdminRoomRelated = UserAdminRoom.IsAdminRoomInvite(ev, this.botUserId)
+                        || ev.room_id === this.config.matrix_admin_room;
+                    if (this.bridgeBlocker?.isBlocked && !isAdminRoomRelated) {
                         log.info(`Bridge is blocked, dropping Matrix event ${ev.event_id} (${ev.room_id})`);
                         return;
                     }
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -1583,7 +1583,11 @@ export class Main {
     }
 
     async disableHookHandler() {
-        await this.slackHookHandler?.close().then(() => log.info("Disabled hook handler"));
+        if (this.slackHookHandler) {
+            await this.slackHookHandler.close();
+            this.slackHookHandler = undefined;
+            log.info("Disabled hook handler");
+        }
     }
 
     enableHookHandler() {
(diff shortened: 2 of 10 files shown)


src/datastore/NedbDatastore.ts
```diff
@@ -63,6 +65,14 @@ export class NedbDatastore implements Datastore {
         private readonly teamStore: NedbDb) {
     }
 
+    storeUserActivity(matrixId: string, activity: UserActivity): Promise<void> {
```
tadzik, line 68: Decided to give up on these completely since NeDB is already deprecated since a few versions ago.

config/config.sample-complete.yaml
```diff
@@ -7,6 +7,9 @@ homeserver:
 
 username_prefix: "slack_"
 
+
+RMAU_limit: 100
```
Half-Shot, line 11: We tend to go lowercase with these. I'm also wondering if we should have co-located it with the room limits, but it's probably a bit late for that now.

src/Main.ts
```diff
@@ -186,6 +214,10 @@ export class Main {
             controller: {
                 onEvent: async(request) => {
                     const ev = request.getData();
+                    if (this.bridgeBlocker?.isBlocked) {
```
Half-Shot, line 217: We might still want to handle admin commands

src/Main.ts
```diff
@@ -1499,4 +1552,30 @@ export class Main {
...
+        await this.slackHookHandler?.close().then(() => log.info("Disabled hook handler"));
+    }
+
+    enableHookHandler() {
+        this.slackHookHandler = new SlackHookHandler(this);
+        log.info("Enabled hook handler");
+    }
+
+    async disableRtm() {
+        await this.slackRtm?.disconnectAll().then(() => log.info("Disabled RTM"));
```
Half-Shot, line 1573: Ooi, why the `then`? This should be identical?
```suggestion
        await this.slackRtm?.disconnectAll();
        log.info("Disabled RTM");
```

src/Main.ts
```diff
@@ -66,8 +66,29 @@ interface MetricsLabels { [labelName: string]: string; }
...
 type TimerFunc = (labels?: Partial<Record<string, string | number>> | undefined) => void;
 
-export class Main {
+class SlackBridgeBlocker extends BridgeBlocker {
+    constructor(userLimit: number, private slackBridge: Main) {
+        super(userLimit);
+    }
+
+    async blockBridge() {
+        await this.slackBridge.disableHookHandler();
```
Half-Shot, line 75: Would add logs here

src/Main.ts
```diff
@@ -66,8 +66,29 @@ interface MetricsLabels { [labelName: string]: string; }
...
+        await this.slackBridge.disableRtm();
+        await super.blockBridge();
+    }
+
+    async unblockBridge() {
+        if (this.slackBridge.config.rtm?.enable) {
```
Half-Shot, line 81: And here

Half-Shot (commented):
Looks good, a few concerns

src/Main.ts
tadzik: The alternative would fire the log.info() even if slackRtm is undefined and thus not really disabled. It does look confusing though, an `if` may be less concise but is probably more obvious. Will change it.

config/config.sample-complete.yaml
tadzik: I thought I was keeping it consistent with something, but apparently it was the exact opposite. Fixed in 121e102

src/Main.ts
tadzik: Done in f085cff

src/Main.ts
tadzik: Done in f085cff

src/Main.ts
tadzik: Done in bc8a22e

src/Main.ts
tadzik: Excluded in 27aa15a

Half-Shot (approved):

PR #613, opened 2021-09-07
Automatically invite users to private rooms when a new message comes in over their RTM stream

Half-Shot:
This uses the bot intent to do a simple state check.


tadzik (approved):
LGTM

PR #621, opened 2021-10-01
Add metric for tracking our success rate with oauth

Half-Shot:
(Also fix a small bug around the blocked metric name)

### DIFF
--- a/src/Main.ts
+++ b/src/Main.ts
@@ -18,7 +18,7 @@ import {
     Bridge, BridgeBlocker, PrometheusMetrics, StateLookup,  StateLookupEvent,
     Logging, Intent, UserMembership, WeakEvent, PresenceEvent,
     AppService, AppServiceRegistration, UserActivityState, UserActivityTracker, UserActivityTrackerConfig } from "matrix-appservice-bridge";
-import { Gauge } from "prom-client";
+import { Gauge, Counter } from "prom-client";
 import * as path from "path";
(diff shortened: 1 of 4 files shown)


src/OAuth2.ts
```diff
@@ -113,6 +117,10 @@ export class OAuth2 {
         // expects inbound roomIds to be 32 chars.
         const token = uuid().substr(0, INTERNAL_ID_LEN);
         this.userTokensWaiting.set(token, userId);
+        setTimeout(() => {
+            this.main.incCounter(METRIC_OAUTH_SESSIONS, {result: "failed", reason: "timeout"});
```
Half-Shot, line 121: I should log something here...

src/Main.ts
```diff
@@ -408,16 +410,22 @@ export class Main {
             name: METRIC_PUPPETS,
         });
         const bridgeBlocked = prometheus.addGauge({
-            name: "bridge_blocked",
+            name: "blocked",
```
tadzik, line 413: Is that the bug? What's wrong with `bridge_blocked`?

Half-Shot: `prometheus.addGauge` prefix's with bridge_ by default, so you end up with bridge_bridge_blocked :p

my bad for not catching the first time.

tadzik: Oh on /o\ Good catch!

tadzik (approved):

PR #628, opened 2021-10-21
Fallback to profile lookup if we don't have membership data for a user+room pair

tadzik:
Inspired by https://github.com/Half-Shot/matrix-appservice-discord/blob/develop/src/matrixeventprocessor.ts#L425, which we should probably integrate in MAB to not duplicate this logic everywhere.

### DIFF
--- a/changelog.d/628.bugfix
+++ b/changelog.d/628.bugfix
@@ -1,1 +1,1 @@
-Make sure we use correct displaynames and avatars on Slack even if we lost track of room state updates
+Improve reliability of Matrix users being bridged with correct displayname and avatar
--- a/changelog.d/628.bugfix
+++ b/changelog.d/628.bugfix
@@ -1,0 +1,1 @@
+Make sure we use correct displaynames and avatars on Slack even if we lost track of room state updates
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -486,7 +486,7 @@ export class BridgedRoom {
         const body: ISlackChatMessagePayload = {
             ...matrixToSlackResult,
             as_user: false,
-            username: user.getDisplaynameForRoom(message.room_id) || matrixToSlackResult.username,
+            username: await user.getDisplaynameForRoom(message.room_id) || matrixToSlackResult.username,
         };
         const text = body.text;
         if (!body.attachments && !text) {
@@ -505,7 +505,7 @@ export class BridgedRoom {
             }
         }
 
-        const avatarUrl = user.getAvatarUrlForRoom(message.room_id);
+        const avatarUrl = await user.getAvatarUrlForRoom(message.room_id);
 
         if (avatarUrl && avatarUrl.indexOf("mxc://") === 0) {
             body.icon_url = this.main.getUrlForMxc(avatarUrl);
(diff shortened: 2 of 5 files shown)


src/Main.ts
```diff
@@ -540,6 +540,10 @@ export class Main {
         return this.stateStorage?.getState(roomId, eventType, stateKey);
     }
 
+    public async getUserProfile(userId: string): Promise<UserProfile|undefined> {
+        return this.botIntent.matrixClient.getUserProfile(userId);
```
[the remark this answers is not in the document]
tadzik: Fixed now in b922085caa3579cac97b387692bb36ba58e087a7. Saves us a whole bunch of extra complexity :)

changelog.d/628.bugfix
```diff
@@ -0,0 +1 @@
+Make sure we use correct displaynames and avatars on Slack even if we lost track of room state updates
```
Half-Shot, line 1: Hm, this feels more like developer talk than something users might understand. Would prefer something like: "Improve reliability of Matrix users being bridged with correct displayname and avatar"

src/BridgedRoom.ts
```diff
@@ -486,7 +486,7 @@ export class BridgedRoom {
         const body: ISlackChatMessagePayload = {
             ...matrixToSlackResult,
             as_user: false,
-            username: user.getDisplaynameForRoom(message.room_id) || matrixToSlackResult.username,
+            username: (await user.getDisplaynameForRoom(message.room_id)) || matrixToSlackResult.username,
```
Half-Shot, line 489: Can this throw horribly, and will we lose the messages?

src/BridgedRoom.ts
```diff
@@ -505,7 +505,7 @@ export class BridgedRoom {
             }
         }
 
-        const avatarUrl = user.getAvatarUrlForRoom(message.room_id);
+        const avatarUrl = await user.getAvatarUrlForRoom(message.room_id);
```
Half-Shot, line 508: same deal

src/Main.ts
```diff
@@ -544,6 +544,10 @@ export class Main {
         return this.stateStorage?.getState(roomId, eventType, stateKey);
     }
 
+    public async getUserProfile(userId: string): Promise<UserProfile|undefined> {
+        return this.botIntent.getProfileInfo(userId);
```
Half-Shot, line 548: Do we not expose botIntent?, this feels like an odd function to have

Half-Shot (changes requested):
A few comments

src/Main.ts
tadzik: Indeed, it needs not exist – though it did make testing a lot easier ;) Removed in bbb79af825dd035d4a4b826389b917a7e2bb106e, and adjusted the tests accordingly.

changelog.d/628.bugfix
tadzik: Good point. Fixed in c921200

src/BridgedRoom.ts
tadzik: Good call. It cannot since 6e1fa00

Half-Shot (approved):

PR #639, opened 2021-11-18
Build docker image against node 16

Half-Shot:


README.md
```diff
@@ -16,7 +16,7 @@ environments.
 
 Hosting this bridge requires you to have a Matrix homeserver. In order to
 connect a Slack Workspace to your bridge, you will need permission to add bots
-to it. You will also need Node.JS 10+ or Docker on your system.
+to it. You will also need Node.JS 12+ or Docker on your system.
```
Half-Shot, line 19: We've always expected Node 12+, but failed to document it.

Dockerfile
```diff
@@ -1,21 +1,21 @@
-FROM node:14-alpine AS BUILD
+FROM node:16-alpine AS BUILD
 
 # git is needed to install Half-Shot/slackdown
 RUN apk add git
 WORKDIR /src
 
 COPY package.json package-lock.json /src/
-RUN npm install
+RUN npm install --ignore-scripts
```
tadzik, line 8: Isn't this a good time to switch it to `npm ci`, so that we actually respect what's in package-lock?

Half-Shot: This is an excellent time for that, how could I forget.

Half-Shot:
Assuming otherwise happiness.

PR #640, opened 2021-11-23
Update dependencies for axios, matrix-appservice-bridge and pgpromise

Half-Shot:
I've also removed winston because we don't directly use this dependency, rather we rely on the upstream logging from m-a-b.


tadzik (approved):
LGTM

PR #646, opened 2021-12-08
Fix cases of nonawaited promises

Half-Shot:
In Node.16, a non-awaited thrown promise kills the process. We don't want that!


tadzik (approved):
Looks good, just needs a newsfile

src/app.ts
```diff
@@ -68,14 +68,13 @@ const cli = new Cli({
...
-            try {
-                await main.killBridge();
-            } catch (ex) {
-                log.warn("Failed to kill bridge, exiting anyway");
-            }
-            process.exit(1);
+            main.killBridge().catch((ex) => {
+                log.warn("Failed to kill bridge, exiting anyway", ex);
+            }).finally(() => {
+                process.exit(0);
```
Half-Shot, line 76: We changed the status here, because 1 is not correct if the bridge managed to exit imo.

daenney: One way you can handle this to give more context is to use a bitmask for possible failure cases, flipping a bit in the mask when you encounter a failure mode during shutdown. Then you take the mask bits, convert it to a number and use that as your exit code. So the process shutting down but having had issues to kill the bridge could end up being a status code of 2, whereas is `killBridge()` didn't fail you'd exit with 0 etc.

tadzik (approved):

src/app.ts
Half-Shot: it feels slightly overkill to bitmask it, given we basically either exit-OK, or exit-something-failed. I can just return a 2 though on failure, reserving 1 for actual crashes.

PR #648, opened 2021-12-11
Fix inverted check if message was sent

ewilderj:
Hey there @Half-Shot , this logic error was activated when the promise stuff was tidied up, and stops the bridge sending messages to the matrix room.


Half-Shot (approved):
Ugh, yeah this shouldn't have happened. Thank you for being awake :coffee:

PR #656, opened 2022-01-16
Support linux/arm64 for Docker deployments

### LINKED ISSUE #655: Add ARM64 support - What is needed to make this happen?
**Is your feature request related to a problem? Please describe.**
I wanted to try this appservice out with my own matrix deployment. I had to realize that there's no official support available for the ARM64 architecture. As more hyperscalers such as Oracle are now providing instances based on the ARM64 architecture I find it important to try to support those architectures as well so people on other architectures can use this appservice as well.

**Describe the solution you'd like**
What is needed to add support for other architectures? Are there any things which make this difficult or which need much work getting it done?

**Describe alternatives you've considered**
Not supporting the architecture at all, besides more users now being on ARM64.

**Additional context**

Half-Shot:
Fixes #655 

I'm hoping this is relatively easy to do.


jaller94 (approved):
The build logs look good,
however, the pipeline duration grew a lot when compared to the latest run on the `develop` branch.

`develop` branch: 57s
This branch: 6m 35s

GoliathLabs:
> > The build logs look good, however, the pipeline duration grew a lot when compared to the latest run on the `develop` branch.
> > `develop` branch: 57s This branch: 6m 35s
> 
> Well spotted. That's a little uncomfortable. I wonder if we should only build arm64 for merges / releases, since we are unlikely to hit an architectural bug

That's what we're doing with some other projects as well. QEMU is not exactly known for its speed when it comes to building stuff for other architectures. So it would be a good idea to run the pipeline only for releases and merges.

PR #665, opened 2022-02-14
Remove uses of the deprecated JavaScript function substr()

jaller94:
`substr()` has been deprecated. For all the remaining uses in this code base `slice()` bahaves the same.

### DIFF
--- a/src/rooms/UserAdminRoom.ts
+++ b/src/rooms/UserAdminRoom.ts
@@ -49,7 +49,7 @@ export class UserAdminRoom {
         }
         const args: string[] = ev.content.body.split(" ");
         const command = args[0].toLowerCase();
-        log.info(`${this.userId} sent admin message ${args[0].substring(32)}`);
+        log.info(`${this.userId} sent admin message ${args[0].slice(32)}`);
         if (command === "help") {
             return this.handleHelp();
         }
--- a/src/rooms/UserAdminRoom.ts
+++ b/src/rooms/UserAdminRoom.ts
@@ -49,7 +49,7 @@ export class UserAdminRoom {
         }
         const args: string[] = ev.content.body.split(" ");
         const command = args[0].toLowerCase();
-        log.info(`${this.userId} sent admin message ${args[0].substr(32)}`);
+        log.info(`${this.userId} sent admin message ${args[0].substring(32)}`);
         if (command === "help") {
             return this.handleHelp();
         }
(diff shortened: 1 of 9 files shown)


src/rooms/UserAdminRoom.ts
```diff
@@ -49,7 +49,7 @@ export class UserAdminRoom {
         }
         const args: string[] = ev.content.body.split(" ");
         const command = args[0].toLowerCase();
-        log.info(`${this.userId} sent admin message ${args[0].substr(32)}`);
+        log.info(`${this.userId} sent admin message ${args[0].substring(32)}`);
```
jaller94, line 52: Why would we skip the first 32 characters of a admin message?
If this is trying to truncate the command to a maximum of 32 characters:
```suggestion
        log.info(`${this.userId} sent admin message ${args[0].length > 32 ? `$args[0].substring(0, 32)}…` : args[0]}`);
```

tadzik: Hah, probably an oversight – unless we're trying to hide credentials in some quirky way.

The ternary is not needed, substr() won't "lenghten" the string if it's shorter:
```
> "foo".substr(0, 32)
'foo'
```

tadzik (commented):
TIL that `substr()` is deprecated. No complaints here :)

src/rooms/UserAdminRoom.ts
jaller94: I added the ternary to append the ellipses so you know that some characters at the end are missi…

jaller94: @Half-Shot You've last touched this code.
Do you remember if skipping the first 32 characters is preferred over limiting the length?

tadzik: > I added the ternary to append the ellipses so you know that some characters at the end are missi…

Ah, indeed, didn't notice that. It might be an off-by-one error if you have 33 characters, but it's no big deal :)

Half-Shot: I can't remember this one, but it sounds like that's what I was trying to achieve.

Half-Shot:
Sad that their isn't an eslint rule for this, looks good to me :)

jaller94:
* Added an ESLint rule
* Changed to `.slice` as it behaves mostly like `.substring` and these sources had good reasons for favouring one:
  * [the ESLint team](https://github.com/eslint/eslint/pull/9010)
  * and [this MasteringJS blogger](https://masteringjs.io/tutorials/fundamentals/substring-vs-slice)

tadzik (approved):

Issue #634, opened 2021-10-26
Support threads with m.thread relation (MSC3440)


jaller94:
Threads are being added to the Matrix protocol soon. Because Slack already supports threads in its protocol and interface, it would be good to bridge that correctly. Currently, Slack threads are bridged as replies.

Related issues:
* Meta issue: https://github.com/vector-im/element-meta/issues/3
* MSC3440: https://github.com/matrix-org/matrix-doc/pull/3440

## Definition of Done
* [x] Slack threads are bridged to Matrix as `m.thread` relations – removing the reply relation we currently set.
* [x] Matrix threads are bridged to Slack as threads.

abmantis:
What would be the behavior in clients that do not implement threads yet? Will they still fallback to replies?

jaller94:
No, the fallback will like be that the messages are shown without context in the main thread.
This is unlikely to affect the plan of implementing this in a bridge once it's in the stable Matrix Spec.

abmantis:
Will this be only implemented when this gets into the stable spec? :(

Half-Shot:
We're planning to implement this sooner.

PR #668, opened 2022-02-22
Support threads in Matrix (MSC3440)

jaller94:
Fixes https://github.com/matrix-org/matrix-appservice-slack/issues/634

Note: Slack doesn't seem to have a separate UI for replies. Everything is a thread.

Screenshot_2022-03-03_16-54-22

## Definition of Done
* [x] Slack threads are bridged to Matrix as `m.thread` relations – removing the reply relation we currently set.


jaller94:
Known issues:

* [ ] Slack's feature "Also send to [main timeline]" isn't supported in Matrix. Messages will only show up in the thread.
* [ ] Sending images and other files does not work in threads.
  * [ ] Slack -> Matrix: The image arrives in the main timeline, not in the thread.
  * [ ] Matrix -> Slack: The message appears in the thread, but is empty.

Half-Shot:
> Slack's feature "Also send to [main timeline]" isn't supported in Matrix. Messages will only show up in the thread.

Can we do a dumb thing where we just send the event twice to Matrix? Would this be too spammy? Alternatively should we comment on the proposal to raise the gap in the spec?

> Sending images and other files does not work in threads. 

I'd be reasonably okay with leaving this as an open issue to fix, if it's not an easy fix.

src/SlackGhost.ts
```diff
@@ -312,6 +316,29 @@ export class SlackGhost {
...
+                "rel_type": "io.element.thread",
+                // If the reply event is part of a thread, continue the thread.
+                // Otherwise, attach a thread to the reply event.
+                "event_id": replyEvent.content["m.relates_to"]?.event_id ?? replyEvent.event_id,
+                "m.in_reply_to": {
+                    event_id: replyEvent.event_id,
+                },
+            },
+            "msgtype": "m.text", // for those who just want to send the reply as-is
+            "body": `${fallbackText}\n\n${this.prepareBody(text)}`,
```
Half-Shot, line 335: So I'm not sure if you still need a fallback or not, looking at what Element does:

image

changelog.d/634.feature
```diff
@@ -0,0 +1 @@
+Support threads with m.thread relation (MSC3440)
```
Half-Shot, line 1:
```suggestion
Support bridging Slack message threads with new m.thread relations (MSC3440)
```

Half-Shot: (wanted to add a bit of colour)

Half-Shot (approved):
Code looks super clean and appears to work, nice one!

src/SlackGhost.ts
jaller94: Element does not provide this fallback.
I guess if a client or bridge (e.g. IRC) wants to display the relation, it up to that cleint's or bridge's implementation to provide an adequate representation.

PR #669, opened 2022-03-03
Fix: Don't create an internal ping room on every restart

jaller94:
Because the parameters were swapped a new ping room is created on every bridge restart.

We don't include a cleanup of false entries as they are no harm and there hasn't been a stable release since this was introduced in PR #666.

Issue description: https://github.com/matrix-org/matrix-appservice-slack/pull/666/files#r818765045


Half-Shot (approved):
Thank you for fixing my cockup!

PR #677, opened 2022-03-21
Make towncrier build the news instead of printing its version

jaller94:
## Before
```
./scripts/towncrier.sh 
towncrier, version 21.9.0
```

## After
```
./scripts/towncrier.sh
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing news fragments...
I want to remove the following files:
/home/jaller94/Git/matrix-appservice-slack/changelog.d/664.misc
…
```

### DIFF
--- a/changelog.d/677.misc
+++ b/changelog.d/677.misc
@@ -1,0 +1,1 @@
+Fix towncrier script for summarising the newsfiles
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
Thanks! Do we need to do this for the other bridge projects?

PR #680, opened 2022-04-20
Explicit that rtm.enabled needs to be set to true

OdyX:
By default, `rtm` is not set and defaults to `false`. Debugging why messages don't land in Matrix from Slack can be _very_ confusing.

### DIFF
--- a/docs/link_channels.md
+++ b/docs/link_channels.md
@@ -58,6 +58,8 @@ The Real Time Messaging (RTM) API is the newer and recommended way to use the br
       ```
       link --channel_id CHANNELID --room !the-matrix:room.id --slack_bot_token xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxxx
       ```
+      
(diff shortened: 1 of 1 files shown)


docs/link_channels.md
```diff
@@ -58,6 +58,8 @@ The Real Time Messaging (RTM) API is the newer and recommended way to use the br
       ```
       link --channel_id CHANNELID --room !the-matrix:room.id --slack_bot_token xoxb-xxxxxxxxxx-xxxxxxxxxxxxxxxxxxxx
       ```
+      
+If messages get sent from Matrix to Slack, but not from Slack to Matrix, make sure that `rtm.enabled` is set to `true` in the `config.yaml` file.
```
Half-Shot, line 62: Hm, I see the intention but technically you can use this bridge without RTM, so we might want to prefer to:

1. Improve our setup documentation to explain why RTM is needed.
2. Add a separate troubleshooting list for folks to go through.

jaller94:
This PR was closed because we deleted the `master` branch.
If you're still interested, please re-open the PR against the `develop` branch.

PR #685, opened 2022-07-15
Move to yarn

Half-Shot:
To be inline with the rest of our projects.

CI will fail due to buildkite not being updated. See #688 for working CI.

PR #688, opened 2022-07-15
Move to GitHub Actions

### LINKED ISSUE #682: Move to GitHub actions for CI
We're meant to be moving away from Buildkite, which just means adapting our existing buildkite pipelines into GitHub workflows.

Half-Shot:
Fixes #682 
~~Depends on #685~~

- The buildkite CI will be failing, but that's fine.
- The rest should pass


.github/workflows/newsfile.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with: # Needed for comparison
+          fetch-depth: 0
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.9'
+      - run: pip install towncrier==21.9.0
+      - name: ":newspaper: Newsfile"
+        run: python3 -m towncrier.check --compare-with=origin/main
```
AndrewFerr, line 19: Should we add `scripts/check-newsfragment` from one of the other bridges (`*-bridge`, `*-irc`, `*-discord`) and run that?

Half-Shot: Done!

scripts/check-newsfragment
```diff
@@ -0,0 +1,46 @@
...
+
+# make sure that origin/develop is up to date
+git remote set-branches --add origin develop
+git fetch -q origin develop
+
+pr="$PULL_REQUEST_NUMBER"
+
+# Print a link to the contributing guide if the user makes a mistake
+CONTRIBUTING_GUIDE_TEXT="!! Please see the contributing guide for help writing your changelog entry:
+https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#making-pull-requests"
```
AndrewFerr, line 19: This has to use the ugly anchor of `#%EF%B8%8F-pull-requests`.  Behold:
https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#making-pull-requests
https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#%EF%B8%8F-pull-requests

scripts/check-newsfragment
```diff
@@ -0,0 +1,46 @@
...
+
+# If check-newsfragment returns a non-zero exit code, print the contributing guide and exit
+python3 -m towncrier.check --compare-with=origin/develop || (echo -e "$CONTRIBUTING_GUIDE_TEXT" >&2 && exit 1)
+
+echo
+echo "--------------------------"
+echo
+
+matched=0
+for f in $(git diff --diff-filter=d --name-only FETCH_HEAD... -- changelog.d); do
```
AndrewFerr, line 29: This needs to exclude `git.keep` (or at least it did when I was testing this script with `matrix-appservice-discord`.
```
for f in $(git diff --diff-filter=d --name-only FETCH_HEAD... -- changelog.d | grep -xv changelog.d/git.keep); do
```

AndrewFerr (changes requested):

scripts/check-newsfragment
Half-Shot: well caught, ugh

scripts/check-newsfragment
Half-Shot: We might be okay, we're using `.gitkeep` and it probably ignores dot files.

AndrewFerr (approved):

PR #696, opened 2022-07-29
Add reactions:write to user scopes

Half-Shot:

### DIFF
--- a/changelog.d/696.bugfix
+++ b/changelog.d/696.bugfix
@@ -1,0 +1,1 @@
+Add missing `reactions:write` scope to oauth scopes.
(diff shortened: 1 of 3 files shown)


tadzik (approved):
Looks okay, but what is it actually changing?

Half-Shot:
Failing to ask for this in previous releases meant we weren't allowing Matrix users to react on Slack (which the upstream ansible-deploy was correcting us on https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bridge-appservice-slack.md)

PR #697, opened 2022-07-29
Drop buildkite badge

Half-Shot:


tadzik (approved):

jaller94:
Buildkite was replaced with Github Actions.
Merging without a newsfile, because I think that removing the badge from the README.md isn't newsworthy.

PR #701, opened 2022-08-11
Update schema v14 -> s/= NULL/IS NULL/

Half-Shot:
We fell into the trap of trying to use a comparison operator for null values.


jaller94 (approved):
Looks good to me.

> [...] the expression `NULL = NULL` returns false. This is because NULL is not equal to any value even itself.

Source: https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-is-null/

Half-Shot:
Always catches me out, because JS treats NULL differently :)

PR #704, opened 2022-08-12
Update matrix-appservice-bridge to 5.1.0

### LINKED ISSUE #702: The /metrics path should not expect auth
This is because addAppservicePath now expects a token, which isn't something we want for metrics.

https://github.com/matrix-org/matrix-appservice-bridge/blob/2dd5a0926d730e2ec9cd7f5a552b1c32e238a860/src/bridge.ts#L957-L990

Half-Shot:
Fixes #702 by upgrading a dependency.

https://github.com/matrix-org/matrix-appservice-bridge/releases/tag/5.1.0


jaller94 (approved):
Looks good to me.

PR #705, opened 2022-08-19
Fix allowing multiple puppets on the same team for the same user

Half-Shot:


jaller94 (changes requested):
The JavaScript and SQL looks to be correct and safe.
I've added some notes about how I think the code and user message could be improved but I think that making a one-off migration message pretty is likely not worth our time.

Ping me on Matrix when you need a re-review. I'm drowning in GitHub notifications.

jaller94 (approved):
Looks ok to me.
The remaining review noted I had made are optional and maybe just personal preference. (e.g. no unnecessary spaces and new lines in the Matrix message.)

PR #709, opened 2022-10-11
Add falling back to threads

### LINKED ISSUE #708: Messages in threads on matrix show replies as well as threads unlike matrix native threads
On element android (and now nehko) messages in threads don't show the reply fallbacks but for some reason the threads made by appservice-slack show both. We must be malforming the events somehow.

Cadair:
This adds the flag to the thread which tells clients the reply is included as a fallback, this should stop slack threads displaying as replies in threads.

fixes #708

### DIFF
--- a/changelog.d/709.bugfix
+++ b/changelog.d/709.bugfix
@@ -1,0 +1,1 @@
+Correctly set "is_falling_back" flag in thread events so that events in threads are not also displayed as replies.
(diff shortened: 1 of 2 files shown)


Cadair:
I have tested this locally and it seems to work :partying_face:

Half-Shot (approved):

PR #715, opened 2022-11-07
Don't track IDE-related files

AndrewFerr:
I subscribe to the philosophy that IDE files should not be checked into version control.  Besides, the VSCode launch config actually made local development more difficult for me (as it was maybe missing a task file or something?), and we shouldn't be burdened with keeping launch configs up-to-date anyhow.


Half-Shot (approved):
Yeah, I think that's reasonable.

PR #717, opened 2022-11-15
Fix admin commands getting stuck 

AndrewFerr:
The two main reasons why admin commands got stuck are:
1. Calling [`yargs.parse`](https://yargs.js.org/docs/#api-reference-parseargs-context-parsecallback) multiple times apparently doesn't behave well when command handlers return a Promise:
> Note: parse() should be called only once when [command()](https://yargs.js.org/docs/#command) is called with a handler returning a promise. If your use case requires parse() to be called several times, any asynchronous operation performed in a command handler should not result in the handler returning a promise.
2. When `yargs.parse` throws, or the callback given to it throws, `yargs.argv` is permanently set to a `YError`, and all future calls to `yargs.parse` throw that same error.

This PR cleans up some of our `yargs` usage & moves all asynchronicity of command callbacks outside of `yargs.parse`.

### DIFF
--- a/changelog.d/717.bugfix
+++ b/changelog.d/717.bugfix
@@ -1,0 +1,1 @@
+Prevent admin commands from being unusable after an invalid command is issued.
--- a/src/AdminCommand.ts
+++ b/src/AdminCommand.ts
@@ -14,39 +14,30 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { Options } from "yargs";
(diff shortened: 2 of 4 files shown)


jaller94 (approved):
Looks good.

AndrewFerr:
The latest commit is needed to fix asynchronous commands not being awaited properly.  It also moves error handling back into `Main.ts` to catch errors more generically & to minimize the size of the diff.

PR #722, opened 2022-11-22
Invite own ghost in DMs if needed

AndrewFerr:
Inviting a Slack ghost to a DM won't invite the ghost for your own Slack user, meaning that messages you send in that DM from Slack can only be bridged if your Slack ghost is in the Matrix DM.  This PR ensures that your Slack ghost gets invited when you send a message in Slack.

### DIFF
--- a/src/BridgedRoom.ts
+++ b/src/BridgedRoom.ts
@@ -954,6 +954,17 @@ export class BridgedRoom {
         // Dedupe across RTM/Event streams
         this.addRecentSlackMessage(message.ts);
 
+        if (this.SlackType === "im") {
+            const intent = await this.getIntentForRoom();
+            if (intent.userId !== ghost.intent.userId &&
+                !(await intent.matrixClient.getRoomMembers(this.matrixRoomId, undefined, ["invite", "join"]))
+                    .map(m => m.membershipFor).includes(ghost.intent.userId))
+            {
+                await intent.invite(this.matrixRoomId, ghost.matrixUserId);
+                await ghost.intent.join(this.matrixRoomId);
+            }
+        }
+
         ghost.bumpATime();
         this.slackATime = Date.now() / 1000;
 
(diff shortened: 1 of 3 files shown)


src/BridgedRoom.ts
```diff
@@ -954,6 +954,17 @@ export class BridgedRoom {
         // Dedupe across RTM/Event streams
         this.addRecentSlackMessage(message.ts);
 
+        if (this.SlackType === "im") {
+            const intent = await this.getIntentForRoom();
+            if (intent.userId !== ghost.intent.userId &&
+                !(await intent.matrixClient.getRoomMembers(this.matrixRoomId, undefined, ["invite", "join"]))
+                    .map(m => m.membershipFor).includes(ghost.intent.userId))
+            {
+                await intent.invite(this.matrixRoomId, ghost.matrixUserId);
```
Half-Shot, line 963: Should we skip inviting if we know the user is invited from the prior call, if the API tells us?

Half-Shot (approved):
Seems reasonable

src/BridgedRoom.ts
AndrewFerr: AFAICT the API tells us by having `Intent.invite` throw an error if the user is joined already!  It also seems to be indistinguishable from a "real" error... So checking room membership in advance might be the only safe way to invite someone.

Unless you meant something else?

Half-Shot: Well it's just you were checking the room membership above, in `getRoomMembers`. Can't we short circuit on join?

AndrewFerr: The tricky part is that what triggers this is seeing an inbound Slack message, which doesn't strongly correspond to a Matrix ghost having joined. So I don't think there's anything before this can be used as a hint to ignore having to send an invite.

Half-Shot: Okay, that makes sense to me. My point was only an optimization if there was an easy path :)

PR #728, opened 2022-11-30
Release 2.0.2

AndrewFerr:
The latest RC is currently deployed in staging.  I've tested the fixes relating to DMs and admin commands.


AndrewFerr:
Hm, this fails the Newsfile check.  If there's a preferred means of getting release commits reviewed & approved, I'm all ears!

Half-Shot:
Looks fine, I think we need to just ignore newsfile checks when the changelog is being updated maybe?

Half-Shot (approved):

PR #729, opened 2022-12-02
Correctly attribute mkarg for past contributions

jaller94:
The original PRs were closed because we deleted the branch which they were filed against.
GitHub didn't allow me to reopen them and point them against the `develop` branch.

In the interest of not spending much time on the changes, I copied the changes into a new PR. This left mkarg without an attribution in the Git history and release notes.

This PR tries to rectify the mistake by re-commiting the changes with a changed `Author` for the Git history and changes to the CHANGELOG.

### DIFF
--- a/changelog.d/729.misc
+++ b/changelog.d/729.misc
@@ -1,0 +1,1 @@
+Correctly attribute mkarg for documentation contributions. We also reverted commits and recommited them with the correct Author value set.
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -75,10 +75,10 @@ ever stuck, you can post a question in the
     `$ yarn start -r -c config/config.yaml -u "http://$HOST:$MATRIX_PORT"`
    or with docker:
    
-   ```sh
-   $ docker run --volume /path/to/config/:/config/ matrixdotorg/matrix-appservice-slack \ 
-      -r -c /config/config.yaml -u "http://$HOST:$MATRIX_PORT" -f /config/slack-registration.yaml
-   ```
+```sh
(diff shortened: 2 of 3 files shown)


docs/getting_started.md
```diff
@@ -99,7 +99,7 @@ ever stuck, you can post a question in the
    `homeserver.yaml`):
    
    ```yaml
-    app_service_config_files:
```
AndrewFerr: Unneeded?

jaller94: I don't remember changing this.
It might have been an auto formatter (which I turned off now), being bothered by the additional space character.

AndrewFerr: Ah, `git blame` shows this as coming from `a0bc9448`.

If you like, you can put the space back in (and maybe squash it with fdddcf7) so that we have 0 diffs against `develop` other than changelogs.

changelog.d/729.misc
```diff
@@ -0,0 +1 @@
+Correctly attribute mkarg for documentation contributions. We also reverted commits and recommited them with the correct Author value set.
```
AndrewFerr, line 1: As long as we're editing `CHANGELOG.md`, this entry can be put in there manually as well.

Plus, if we're using this PR to update the latest release, we shouldn't keep any ` changelog.d/` files around, as the typical release flow has them get removed by the towncrier script.

AndrewFerr: We've agreed to not rewrite release history, so this changelog entry can stay & get handled when the towncrier script cuts the next `CHANGELOG.md` update.

AndrewFerr (approved):

PR #733, opened 2023-02-08
Update Provisioning API

justinbot:
Makes some significant updates to the Provisioning API implementation:
* Refactored to use the `ProvisioningApi` class from `matrix-appservice-bridge`. (Which also adds support for authenticating with widgets).
* Added request validation using schemas
* Improved error handling and added consistent error responses

### DIFF
--- a/config/config.sample.yaml
+++ b/config/config.sample.yaml
@@ -156,27 +156,30 @@ slack_hook_port: 9898
 
 # Optional.
 #
-# provisioning:
-  # Whether to enable the provisioning API. Defaults to true.
+#provisioning:
+#  # Whether to enable the provisioning API. Defaults to true.
 #  enabled: true
-
-  # Run a separate HTTP listener for provisioning requests. Optional, defaults to the appservice listener.
+#
+#  # Whether to enable hosting the setup widget page. Defaults to false.
+#  widget: false
+#
+#  # Run a separate HTTP listener for provisioning requests. Optional.
 #  http:
 #    # The port to listen on.
 #    port: 5859
 #    # The host to listen on. Defaults to 0.0.0.0.
 #    host: 0.0.0.0
-
-  # A secure secret token used when making provisioning requests. Defaults to the appservice hs_token.
+#
+#  # A secure secret token used when making provisioning requests. Defaults to the appservice hs_token.
 #  secret: "<GENERATE ME>"
-
-  # Whether to apply rate limiting to provisioning requests. Defaults to true.
+#
+#  # Whether to apply rate limiting to provisioning requests. Defaults to true.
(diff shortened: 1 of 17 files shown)


src/datastore/NedbDatastore.ts
```diff
@@ -23,7 +23,7 @@ import {
     StoredEvent,
     StoredEventDoc,
     UserActivity,
-    UserActivitySet
+    UserActivitySet, ProvisionSession
```
Half-Shot:
```suggestion
    UserActivitySet,
    ProvisionSession
```

config/config.sample.yaml
```diff
@@ -156,13 +156,55 @@ slack_hook_port: 9898
 
 # Optional.
 #
-# provisioning:
-#   enabled: true
+#provisioning:
```
Half-Shot, line 159: I think we have a `config.complete.yaml` too.

src/provisioning/Provisioner.ts
```diff
@@ -0,0 +1,568 @@
+/*
+Copyright 2019 The Matrix.org Foundation C.I.C.
```
Half-Shot: Copyright header needs to be recent :)

src/provisioning/Provisioner.ts
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    public async start(): Promise<void> {
+        if (this.config.http) {
+            await super.start(this.config.http.port, this.config.http.host);
+        }
+        log.info("Provisioning API ready");
+    }
+
+    public fakeRequest(method: string, req: ProvisioningRequest, res: Response) {
```
Half-Shot, line 160: Nit: Would be good to have a docstring to know what this is used for.

src/provisioning/Provisioner.ts
```diff
@@ -0,0 +1,568 @@
...
+                    } else {
+                        req.log.error("Unknown error:", e);
+                        // Send a generic error
+                        const err = new ApiError("An internal error occurred");
+                        err.apply(res);
+                    }
+                }
+            };
+
+        this.addRoute("post", "/getbotid", wrapHandler(this.getBotId), "getBotId");
```
Half-Shot, line 134: Can we rename this endpoint to /getbotinfo (with perhaps a `/getbotid` fallback). The naming is terrible.

Half-Shot (approved):
Looks good, only a few nits!

src/provisioning/Provisioner.ts
justinbot: Honestly the whole API definition needs to be redone (why is _everything_ a `POST` request?). I would say let's leave it for then.

PR #734, opened 2023-02-08
Add new setup widget UI

justinbot:
Following #733 which adds support for widgets authenticating with the Provisioning API, this PR adds a new widget UI to manage workspaces and linked channels from a room:

Screenshot 2023-02-08 at 2 52 13 PM

Screenshot 2023-02-08 at 2 53 05 PM

https://user-images.githubusercontent.com/20007954/217835301-52949561-bd90-40ff-a110-7a74eb82932f.mov

PR #735, opened 2023-02-13
Fix lint errors and run CI on all pull requests

justinbot:
Some lint errors were missed in #734 due to CI only running on PRs against `develop`.
(#734 was against another base branch, and CI does not run again when the base branch is changed)

This fixes the lint errors, and configures CI to run against _all_ PRs.


jaller94 (approved):
Looks good to me.

changelog.d/735.misc
```diff
@@ -0,0 +1 @@
+Fix lint errors and run CI on all PRs.
```
Half-Shot: N.B. if this PR is fixing an issue with another PR, it probably doesn't need to appear in the changelog. What you can do is paste the same changelog you gave to the previous PR into this file, and it will automerge at release time.

Half-Shot (changes requested):
One nit RE changelogs.

Half-Shot (approved):
Looks fine to me, thanks for cleaning this up :+1:

PR #739, opened 2023-03-07
Fix provisioning API base URL in setup widget

justinbot:
Fixes how the setup widget determines the API base URL to use, in order to support hosting on different paths.
e.g. `https://example.com/slack/_matrix/provision/v1/static` -> `https://example.com/slack/_matrix/provision`

This takes advantage of the widget always being hosted at `/v1/static`, so we can just strip that to get the base URL:
https://github.com/matrix-org/matrix-appservice-bridge/blob/7ecbe69fb86e92851af3946135867357e8866743/src/provisioning/api.ts#L166-L167

### DIFF
--- a/widget/src/ProvisioningApp.tsx
+++ b/widget/src/ProvisioningApp.tsx
@@ -25,16 +25,13 @@ export const useProvisioningContext = (): ProvisioningContext => {
 };
 
 /**
- * @param apiPrefix Base path for API requests.
  * @param tokenName Name to use for the session token in localstorage.
  * @param children
  * @constructor
  */
 export const ProvisioningApp: React.FC<React.PropsWithChildren<{
-    apiPrefix: string,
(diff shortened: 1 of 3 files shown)


widget/src/ProvisioningApp.tsx
```diff
@@ -25,16 +25,13 @@ export const useProvisioningContext = (): ProvisioningContext => {
 };
 
 /**
- * @param apiPrefix Base path for API requests.
  * @param tokenName Name to use for the session token in localstorage.
  * @param children
  * @constructor
  */
 export const ProvisioningApp: React.FC<React.PropsWithChildren<{
-    apiPrefix: string,
```
justinbot, line 34: We don't need to know this anymore because instead of constructing the base URL, we're just removing the last bits.

Half-Shot (approved):

PR #752, opened 2023-05-02
Create a new ping room if the previous room cannot be reached

Half-Shot:
If the room the bridge is using to "ping" the homeserver has been deleted, we will fail to start because the pings will always immediately fail. Instead, try to join the room and if we fail, create another. Joins will noop if we're already in the room.

### DIFF
--- a/changelog.d/752.bugfix
+++ b/changelog.d/752.bugfix
@@ -1,0 +1,1 @@
+Fix a case where the bridge will fail to start if it cannot access it's internal ping room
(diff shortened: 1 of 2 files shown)


tadzik:
Fixes https://github.com/matrix-org/matrix-appservice-slack/issues/764, it seems

tadzik (approved):

Half-Shot:
Hey there people on the tail end of 2025, closing this PR as development on the Slack bridge has been wound down now that the foundation is no longer running an instance. I'll work with the foundation on how we might want to communicate the status of this repository, but for now I'll close out any PRs that will not be landing.

PR #775, opened 2024-03-26
Update dependencies, require node 20

tadzik:
Signed off by: Tadeusz „tadzik” Sośnierz \<signoff at tadzik dot net\>


Cadair:
Looks sane to me, but the postgres integration test fail seems legit?

edit: oh wait, looks like it's just busted, so probably not related to this PR.

tadzik:
It's worth considering updating our matrix-appservice-bridge here and dropping NeDB while at it – we'd need to support a DB migration though.

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
    runs-on: ubuntu-latest
```
Half-Shot, line 38: Ideally we should make sure this works before merging.

src/substitutions.ts
```diff
@@ -25,14 +25,6 @@ const log = new Logger("substitutions");
 const ATTACHMENT_TYPES = ["m.audio", "m.video", "m.file", "m.image"];
 const PILL_REGEX = /<a href="https:\/\/matrix\.to\/#\/(#|@|\+)([^"]+)">([^<]+)<\/a>/g;
 
-/**
```
Half-Shot, line 28: Assuming this is now part of the functionality.

Half-Shot (changes requested):
What's needed to make that test pass?

src/substitutions.ts
tadzik: Correct. I've now added tests to verify this.

tadzik:
> What's needed to make that test pass?

It needed the default timeout increased, 2 seconds was not enough to apply DB migrations. Should be good now.

changelog.d/775.misc
```diff
@@ -0,0 +1 @@
+Require node 20, update dependencies.
```
Half-Shot: This needs to be a `.removal`, as we're removing support. I'd put in stronger wording to say that Node 16 and 18 are no longer supported, and users need to update to Node 20.

I wouldn't both stating that dependencies are updated, it doesn't help the user any to know that.

.github/workflows/lint.yml
```diff
@@ -13,6 +13,6 @@ jobs:
     - name: Use Node.js
```
Half-Shot, line 13: N.B. Remember to update the Dockerfile.

Half-Shot (changes requested):

changelog.d/775.misc
tadzik: Good points. Changed in 7f3cdecb454a35f78c6ecb818026723c0f520c99

.github/workflows/lint.yml
tadzik: Done in af9075b3d164750e103b1d0aa221aa09b00b4405

PR #776, opened 2024-03-28
Verify Slack webhook tokens

tadzik:
This will unfortunately require relinking all existing channels that use webhooks.

I pondered making this optional, but in the spirit of being secure by default, it's probably best if we mandate this.

Existing rooms will have to be unlinked and linked again (or have their database entries updated manually).

This also increases SlackHookHandler's test coverage by ~18% (from 9.39% to 27.51%) :)


src/SlackHookHandler.ts
```diff
@@ -85,7 +85,7 @@ export class SlackHookHandler extends BaseSlackHandler {
         }
     }
 
-    private onRequest(req: IncomingMessage, res: ServerResponse) {
+    public _onRequest(req: IncomingMessage, res: ServerResponse) {
```
Half-Shot: This seems quite cheeky. If you're mocking stuff, why not `hookHandler["onRequest"]` to get around the public/private check.

src/SlackHookHandler.ts
```diff
@@ -67,7 +67,7 @@ export class SlackHookHandler extends BaseSlackHandler {
             createServer = (cb) => httpsCreate(tlsOptions, cb);
         }
         return new Promise<void>((resolve, reject) => {
-            const srv = createServer(this.onRequest.bind(this));
+            const srv = createServer(this._onRequest.bind(this));
```
Half-Shot: Not a huge fan of the underscore prefix, smells like snakes

tests/integration/WebhookTest.ts
```diff
@@ -0,0 +1,117 @@
...
+        return checkResult(req, res => {
+            expect(res.statusCode).to.equal(200);
+        });
+    });
+
+    it("will reject webhooks not containing a valid token", () => {
+        let room = new BridgedRoom(harness.main as unknown as Main, {
+            matrix_room_id: '!foo:bar.baz',
+            inbound_id: randomstring.generate(32),
+            slack_webhook_token: randomstring.generate(24),
```
Half-Shot: What about no token at all :). Or null, or `true` etc etc :)

Half-Shot (changes requested):

src/SlackHookHandler.ts
tadzik: Getting around the check seems dirtier to me, to be honest. I won't fight for it though, I'm fine with either.

src/SlackHookHandler.ts
tadzik: Changed in https://github.com/matrix-org/matrix-appservice-slack/commit/2d97794449e3a926a93bcfaf4086874ffb77bc0d?diff=unified&w=1

tests/integration/WebhookTest.ts
tadzik: Added a few more in https://github.com/tadzik/matrix-appservice-slack/pull/776/commits/4d312a35003fecefc339fed6772a0eabf42ec8f7?diff=unified&w=1 :)

changelog.d/776.bugfix
```diff
@@ -0,0 +1 @@
+Verify Slack webhook tokens.
```
tadzik, line 1: I figured `bugfix` is the closest this comes to, since it was arguably a bug that we never did this before.

PR #777, opened 2024-03-28
Add a way to measure test coverage

tadzik:
We had some leftovers of it in .gitignore, now it's actually available as a command


Cadair (approved):
I wont pretend to be able to fully understand this incantation, but if it works then horray.

package.json
```diff
@@ -85,6 +86,8 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "js-yaml": "^4.1.0",
     "mocha": "^10.0.0",
+    "node-mocks-http": "^1.14.1",
```
Half-Shot: hello?

Half-Shot (changes requested):
Changelog please :)

package.json
tadzik: [Is it me you're looking for?](https://youtu.be/Cwnrsd5so9M?t=27)

(Removed in dc9bc72b146c931b6447f386b8aa844e5ebec073, thanks!)

PR #778, opened 2024-03-28
Improve plaintext formatting of Matrix messages sent via webhooks

tadzik:
In webhook mode sending messages comes with some limitations,
including us not being able to impersonate users or upload attachements.
This makes us fallback to plaintext relaybot behaviour,
while preserving as much from the original message as possible.

Signed-off-by: Tadeusz „tadzik” Sośnierz <tadeusz@sosnierz.com>

Depends on: https://github.com/matrix-org/matrix-appservice-slack/pull/776


src/SlackHookHandler.ts
```diff
@@ -234,6 +234,16 @@ export class SlackHookHandler extends BaseSlackHandler {
             return;
         }
 
+        if (params.token !== room.SlackWebhookToken) {
```
Half-Shot, line 237: Do we permit it if no `room.SlackWebhookToken` exists?

Half-Shot (changes requested):
Looks good, one thought. Also needs a changelog.

src/SlackHookHandler.ts
tadzik: Is there any situation when this could be true for webhook-bridged rooms? I don't think the bridge will allow you to set them up that way.

Half-Shot (approved):

PR #800, opened 2026-01-21
Fix newsfile workflow

AndrewFerr:
This updates the versions of the setup-python action & towncrier to match what's used by matrix-appservice-bridge, and to set the PR number in the checker script as m-a-b does.

This was done to fix the workflow failing on recent runs.


jaywink (approved):
The repository is being archived though :)

AndrewFerr:
> The repository is being archived though :)

Yes, just wanted to clean this up in case we want to cut a final release.

